PR #13, opened 2021-11-25
Patch chart, docker and oprerator versions in pull-requests

razvan:
... and some other teaks.

## Goal

Publish PR artifacts with their own version. These artifacts are published in their respective repositories but the source code is NOT tagged in any way.

The following artifacts are published with the exact same PR-version:
- Operator (not published separately but bundled in the docker image)
- Operator Docker image
- Helm chart

## How it works

- patch versions in Cargo.toml using the `cargo-version.py` script.
- build, tag and publish docker image with `make docker `which uses the Cargo.toml version
- package and publish the Helm chart using` make compile-chart` which also uses the Cargo.toml version.

The result is that when creating a PR, the same version is used for all three artefacts.

## TODO

- [x] What Version to use for patching ? Currently "mr-<pr-number>" is used which is not valid SemVer.
- [x] Test this with only one operator first (spark)
- [x] Create a follow-up PR where all operator repositories are patch.


backstreetkiwi (commented):
I highly agree with the goals of this PR, though I cannot give a valid feedback on the implementation.
I disagree with the suffix "mr-*" not being a valid SemVer version. According to https://semver.org/#spec-item-9, we can very well name a "pre-release-version" like "mr-123" or "nightly".

maltesander (approved):
LGTM.

PR #14, opened 2021-11-26
Generate manifests and added kustomization.yaml

fhennig:
This pull request extends the 'helm' job in the 'publish_dev_artifacts' github action to also generate manifest files from the helm chart. It then commits those to the repository.

Also a `kustomization.yaml` was added to support `kubectl kustomize` with the manfests.

The https://github.com/stackable-bot stackable bot is used to make the automatic commit.

Testing has been done here: https://github.com/stackabletech/zookeeper-operator/pull/278


template/.github/workflows/publish_dev_artifacts.yml.j2
```diff
@@ -45,4 +46,23 @@ jobs:
           -u 'github:${{ secrets.NEXUS_PASSWORD }}'
           --upload-file "./$(find  target/helm/ -name *.tgz)"
           "${{ env.REPO_HELM_DEV_URL }}/"
-
+      - name: Clean old manifests
+        run: mkdir -p deploy/manifests && rm -r $(find deploy/manifests -maxdepth 1 -mindepth 1 -not -name kustomization.yaml)
```
razvan, line 50: This will fail if the folder deploy/manifests exists.

razvan (commented):
I recommend you encapsulate all manifest management steps in a script that can be tested locally and maybe also used in Jenkins.

template/.github/workflows/publish_dev_artifacts.yml.j2
fhennig: I don't see how, `mkdir -p` doesn't fail if the directory already exists. Or did you mean a different part of the command?

fhennig: Ah `rm -r` can file if the find command doesn't return anything! I'll fix that

razvan (approved):
lgtm

PR #19, opened 2021-11-26
Stacky McStackface is the new Sönke

lfrancke:


fhennig:
You can use that:

          author_name: Stacky McStackface
          author_email: stackable-bot@users.noreply.github.com

and it will also link to the bots profile and add the picture (not sure if it works with github@stackable.de as well)

teozkr (approved):

teozkr:
github@ should work as well, as long as it's one of the emails registered to the account.

lfrancke:
Thanks both. Yes, the address is registered to that account.

PR #27, opened 2021-11-30
Removed buggy echo line

fhennig:
The `echo` command caused escaped characters to not be escaped anymore in the configmaps.
I removed it. It was adding the "this is a generated file"-notice, which was there twice actually.


razvan (approved):
It looks like the generated manifests (except the crds.yaml) preserve the original template header anyway.

fhennig:
indeed

PR #32, opened 2021-12-13
Added an environment variable to disable lookups in log4j strings.

soenkeliebau:
This is in order to protect against Log4Shell (CVE-2021-44228).


teozkr (approved):
bors r+

PR #33, opened 2021-12-15
Fix broken makefile, adapt K8s manifests for STFU rework

teozkr:


template/Makefile.j2
```diff
@@ -22,8 +22,7 @@ chart-clean:
 	rm -rf deploy/helm/{[ operator.product_string }]-operator/templates/crds.yaml
 
 version:
-	yq eval -i '.version = ${VERSION} | .appVersion = ${VERSION}' deploy/helm/{[ operator.product_string }]-operator/Chart.yaml
-
+	yq -yi '.version = ${VERSION} | .appVersion = ${VERSION}' deploy/helm/zookeeper-operator/Chart.yaml
```
fhennig, line 25: The "zookeeper-operator" looks like it might be a mistake here

template/Makefile.j2
```diff
@@ -34,7 +33,7 @@ crds: deploy/helm/{[ operator.product_string }]-operator/crds/crds.yaml
 
 deploy/helm/{[ operator.product_string }]-operator/crds/crds.yaml:
 	mkdir -p deploy/helm/{[ operator.product_string }]-operator/crds
-	cat deploy/crd/*.yaml | yq e '.metadata.annotations["helm.sh/resource-policy"]="keep"' - > ${@}
+	cat deploy/crd/*.yaml | yq '.metadata.annotations["helm.sh/resource-policy"]="keep"' - > ${@}
```
fhennig, line 36: What is that doing? Is "eval" the default action? I'd be in favor of being explicit in scripts, because it's easier to read

template/Makefile.j2
teozkr: Good point. WIll fix.

template/Makefile.j2
teozkr: I'm starting to suspect that this is down to using two completely different scripts that are just both called `yq`. The one I'm using (Arch's `yq` package, seems to be https://github.com/kislyuk/yq) doesn't have subcommands at all, just like `jq` doesn't. `nixpkgs#yq` also seems to point towards the same source repository. Which one are you using?

template/Makefile.j2
teozkr: fa1585968b1ad9bbd165640c3666711a48a99e6d

template/Makefile.j2
fhennig: Oh! I'm using this one https://github.com/mikefarah/yq it's also the one that's used in the github actions. We should definitely pick one :smile:

razvan (commented):
Not directly related to this PR but probably can be fixed here is the problem with this line in the Makefile:

template/Makefile.j2
razvan: I'm using whatever `snap install yq` installed and I wrote the first version. Sorry.

fhennig: That is also the version I'm using, it references the github repo I linked as well

fhennig: I think since this is also running in the github actions, the mikefarah variant is probably the one we want to code against

teozkr:
Oh, looks like ZooKeeper now calls its binary crate `stackable-zookeeper-operator`. I think that makes sense to roll out, since there is no longer a separate binary crate. Originally I rolled `operator-binary` into `operator` (rather than the reverse), but had to switch that back due to scripting that assumed the former.

template/Makefile.j2
teozkr: Looks like that one is packaged by Arch as `go-yq`, and Nixpkgs as `yq-go`, fwiw. Whichever one we end up using, it should probably also be documented somewhere.

teozkr: f8f71dc298962ae78264662209997c092bb4b460

fhennig (approved):
Assuming you tested this, looks good to me! @razvan you wanna have a look too?

fhennig:
The "VERSION" bit runs fine for me in the zookeeper operator, the crate name is correct although it is in the "...-binary" directory"

razvan (dismissed):
lgtm

template/deploy/helm/[[product]]-operator/templates/deployment.yaml
```diff
@@ -34,6 +34,11 @@ spec:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
         - name: {{ include "operator.appname" . }}
+{[% if operator.product_string in ['zookeeper']%}]
+# Reworked ("STFU") operators require a run argument
+          args:
+            - run
+{[% endif %}]
```
Comment (siegfriedweber) [inline template/deploy/helm/[[product]]-operator/templates/deployment.yaml:41]: * deployment.yaml is not a Jinja2 template.
* The `run` argument could also be added to the Dockerfile. The Docker container would then perform the "default" operation when started and not just complain about missing arguments.
* I would not make the argument conditional. The generated pull request should then be merged when the rework is done. Admittedly, updates on operators which are not reworked yet would not be possible.

siegfriedweber (changes requested):

template/deploy/helm/[[product]]-operator/templates/deployment.yaml
Comment (teozkr) [reply template/deploy/helm/[[product]]-operator/templates/deployment.yaml]: 300fdd60d75d0dfade47b4f81ec6866ab0b6721a

template/deploy/helm/[[product]]-operator/templates/roles.yaml.j2
```diff
@@ -73,6 +73,18 @@ rules:
     resources:
       - pods
       - configmaps
+      - services
+      - endpoints
+    verbs:
```
Comment (siegfriedweber) [inline template/deploy/helm/[[product]]-operator/templates/roles.yaml.j2:78]: The verb `patch` must be added to `services`, `statefulsets`, and `jobs`.
The verbs should be ordered alphabetically.

siegfriedweber (changes requested):

template/deploy/helm/[[product]]-operator/templates/roles.yaml.j2
Comment (teozkr) [reply template/deploy/helm/[[product]]-operator/templates/roles.yaml.j2:78]: 9c0c974ba6b9192d05001d4a2b3df7c5cbc6ce40

siegfriedweber (approved):
Tested successfully with the superset-operator.

https://github.com/stackabletech/superset-operator/commit/0d0be251d1e078b39de0a1c7b4d44096439d966e

https://ci.stackable.tech/view/02%20Component%20Tests%20(flex)/job/Superset%20Operator%20Integration%20Tests%20(flex)/8/

PR #45, opened 2021-12-23
Add .dockerignore to managed files and fix typo with retired files

lfrancke:
Diagnosed and found together with @maltesander


maltesander (approved):
LGTM.

PR #50, opened 2022-01-13
Resource names in RBAC rules must be pluralized

teozkr:


teozkr:
Manually applied to https://github.com/stackabletech/kafka-operator/pull/275/commits/da409e62b56b32e2fc669b32df6b39504ad54c89 for now, but should be rolled into the templating too, so it doesn't get overwritten...

maltesander (approved):
LGTM.

PR #51, opened 2022-01-18
Add reference to Kafka broker clusterrole

teozkr:
Required by https://github.com/stackabletech/kafka-operator/pull/275

Feels like this is the completely wrong place for this kind of thing, ideally we'd have some way for the repository to declare this kind of thing locally.. :/


maltesander (approved):
LGTM.

PR #53, opened 2022-01-18
Various changes

lfrancke:
- YAML formatting (yamllint)
- Pylint
- Pull Requests to separate Helm Repo (helm-test) which also includes renaming them from mr* to pr* again
- Rustdoc will now error on warnings
- Enable bors squash merge again
- Enable helm chart testing and linting


repositories.yaml
```diff
@@ -18,7 +19,7 @@ repositories:
     url: stackabletech/regorule-operator.git
     product_string: regorule
     pretty_string: RegoRule
-    include_productconfig: no
+    include_productconfig: true
```
teozkr, line 22: Typo?

template/deploy/helm/[[product]]-operator/Chart.yaml.j2
```diff
@@ -1,33 +1,11 @@
-# =============
-# This file is automatically generated from the templates in stackabletech/operator-templating
-# DON'T MANUALLY EDIT THIS FILE
-# =============
```
Comment (teozkr) [inline template/deploy/helm/[[product]]-operator/Chart.yaml.j2:4]: Don't we want this header still?

repositories.yaml
lfrancke: Thank you for catching that!

template/deploy/helm/[[product]]-operator/Chart.yaml.j2
Comment (lfrancke) [reply template/deploy/helm/[[product]]-operator/Chart.yaml.j2:4]: I believe the answer is "no" as it'll end up in the generated helm charts and it's more confusing than anything else there IMO

Comment (teozkr) [reply template/deploy/helm/[[product]]-operator/Chart.yaml.j2:4]: Ok

PR #55, opened 2022-01-26
Enable templating for secret operator

teozkr:
- [x] Enable templating for secret-operator
- [x] Fix -operator-operator in templates
- [x] Fix -operator-operator in folder names
- [x] Deploy as DaemonSet rather than Deployment
- [x] Set `CSI_ENDPOINT` env variable
- [x] CRD harmonization


teozkr:
The remaining question here is harmonizing CRD YAML generation. The Makefile currently assumes that the operator's build.rs script will write the CRD YAML, which secret-op currently doesn't because Nix doesn't let it write outside of its build sandbox.

We could make this conditional ("only write the CRD when built outside of Nix"), but I'm disinclined to let them diverge too much, and this kind of code mutation is discouraged upstream (rule 1 of https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script is "Scripts should not modify any files outside of that [$OUT_DIR] directory.").

An alternative used by secret-op is to use the output of `cargo run crd`. This is already supported by all operators, but is somewhat slower (since it causes the check-manifests task to also do a cargo build). Most of this could be mitigated with caching, however.

teozkr:
After (so-far) unanimous agreement in Slack (https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1647251421170359) I'm burning `deploy/crd` entirely.

repositories.yaml
```diff
@@ -50,6 +50,12 @@ repositories:
...
     pretty_string: Apache Druid
+  - name: secret-operator
+    url: stackabletech/secret-operator.git
+    product_string: secret-operator
+    pretty_string: Stackable Secret Operator
+    run_as: custom
+    include_productconfig: false
 
 # Anything specified in the retired_files variable will be deleted.
 # This is uncommented as I had issues with everything being deleted when this was just present as an empty key.
```
soenkeliebau, line 61: We need to specify the 'deploy/crd' folder here I think, otherwise the files in there will remain in the repos as orphans.

soenkeliebau, line 61: We need to specify the 'deploy/crd' folder here I think, otherwise the files in there will remain in the repos as orphans.

soenkeliebau (changes requested):
just one comment, looks good otherwise I think.

I've performed a test run, but not cretated prs to run ci. We can take a look at those once this is merged and the prs are created I think.

repositories.yaml
teozkr: Those files are created by the build.rs scripts (which aren't templated), so they'll just be recreated on the next build.

teozkr: Dupe of https://github.com/stackabletech/operator-templating/pull/55#discussion_r827876855.

soenkeliebau: oh man, sorry for being so absent minded

soenkeliebau (approved):

sbernauer:
FYI @teozkr this only works with non-snap version of yq. Snap version is silly and throws (file exists though)

Only FYI, fine for me ;)
```
make regenerate-charts
rm -rf deploy/helm/superset-operator/configs
rm -rf deploy/helm/superset-operator/crds
mkdir -p deploy/manifests
rm -rf $(find deploy/manifests -maxdepth 1 -mindepth 1 -not -name Kustomization)
yq eval -i '.version = "0.4.0-nightly" | .appVersion = "0.4.0-nightly"' deploy/helm/superset-operator/Chart.yaml
Error: stat deploy/helm/superset-operator/Chart.yaml: no such file or directory
make: *** [Makefile:39: version] Error 1

sbernauer:
See https://github.com/mikefarah/yq/#snap-notes

PR #56, opened 2022-01-26
Remove helm tests for now

lfrancke:
They were not using the Docker image from the current run and cause more trouble


razvan (approved):
lgtm

PR #57, opened 2022-01-31
Allow operators access to Secret objects

teozkr:
Fixes https://github.com/stackabletech/trino-operator/pull/97

/cc @sbernauer @soenkeliebau


maltesander (approved):
LGTM.

PR #59, opened 2022-02-01
Updated PR Template to include a check of the helm chart

fhennig:
I kept forgetting this, but it is important. The operator deployed as a helm chart should work. At the moment this is also the easiest (only?) way to check if the RBAC rules work properly.

Also `make compile-chart` now always rebuilds the charts.


razvan (approved):
lgtm

fhennig:
Looks good to me as well

PR #61, opened 2022-02-04
Overhaul of CI infrastructure

soenkeliebau:
This is a overhaul of our overall CI structure to build, test and deploy artifacts. It moves all build jobs to a single workflow so we can properly declare dependencies between them.

Also removes the Generate Manifests action which adds a commit to all PRs. This is replaced by a job which regenerates charts and manifests and then fails the workflow if `git diff` shows any changes. This is based on the assumption that in that case obsolete crds, charts and/or manifests were committed.

fixes https://github.com/stackabletech/issues/issues/156

Also adds various fixes:

- update helm & cargo deny actions
- remove dependabot and add renovate
- change bors to wait for an approving review
- remove (but not delete) helm roles.yaml so those can be managed in the operators
- fix shebang in generate manifests
- fix weird cp behavior for product config in makefile (subdirectory 'config-spec' should be removed now)
- add reviewdog workflow to run various linters on pull requests
- add udeps job to find unused dependencies

I have created an example PR that should match what will be generated by this once it is merged here: https://github.com/stackabletech/zookeeper-operator/pull/388

PR #62, opened 2022-02-07
Make copying of config-spec conditional

soenkeliebau:
Some operators (looking at you, Regorule) do not have config-spec files, which caused the generate manifest step to fail after PR #61 which changed the syntax for copying these files.

This adds a check if any config-spec files are present and skips copying them if not. This requires that operators do not have .dummy files in place, as was the case for regorule (fixed in https://github.com/stackabletech/regorule-operator/pull/195 )

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -40,7 +40,7 @@ version:
 
 config:
 	if [ -d "deploy/config-spec/" ]; then\
-		mkdir -p deploy/helm/zookeeper-operator/configs;\
+		mkdir -p deploy/helm/{[ operator.product_string }]-operator/configs;\
 		cp -r deploy/config-spec/* deploy/helm/{[ operator.product_string }]-operator/configs;\
 	fi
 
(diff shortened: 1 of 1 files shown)


template/Makefile.j2
```diff
@@ -39,8 +39,10 @@ version:
 	yq eval -i '.version = ${VERSION} | .appVersion = ${VERSION}' deploy/helm/{[ operator.product_string }]-operator/Chart.yaml
 
 config:
-	mkdir -p deploy/helm/zookeeper-operator/configs
-	cp -r deploy/config-spec/* deploy/helm/{[ operator.product_string }]-operator/configs
+	if [ -d "deploy/config-spec/" ]; then\
+		mkdir -p deploy/helm/zookeeper-operator/configs;\
```
lfrancke, line 43: This seems like it was wrong before and is still wrong now :)
Didn't catch the hardcoded reference to ZK before so this can't really have worked....

lfrancke (changes requested):

template/Makefile.j2
soenkeliebau: It worked since the error is in the portion that gets skipped :)
Wouldn't have worked for the other case, no.

lfrancke (approved):

PR #63, opened 2022-02-07
Change bors required checks from publishing to chart testing

soenkeliebau:
Publishing Artifacts step is skipped for bors staging branches but marked as required.

This keeps bors from merging any prs. Since we will move building charts and docker images to earlier stages soon anyway the important steps will all have happened when chart testing is done, so we can make this the required step instead.


maltesander (approved):
LGTM.

PR #65, opened 2022-02-09
Pin action versions

soenkeliebau:
This makes the changes Renovate tried to perform downstream in the repositories to pin action versions.

Added yamllint config to allow less whitespaces before inline comments to accomodate renovate.

Added comments to suppress detect-secret warnings.

### DIFF
--- a/template/.yamllint.yaml
+++ b/template/.yamllint.yaml
@@ -8,3 +8,5 @@ rules:
   line-length: disable
   truthy:
     check-keys: false
+  comments:
+    min-spaces-from-content: 1  # Needed due to https://github.com/adrienverge/yamllint/issues/443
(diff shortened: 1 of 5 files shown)


template/.yamllint.yaml
```diff
@@ -8,3 +8,5 @@ rules:
   line-length: disable
   truthy:
     check-keys: false
+  comments:
+    min-spaces-from-content: 1  # Needed due to https://github.com/adrienverge/yamllint/issues/443
```
lfrancke, line 12: This needs to be done in the main yamllint file as well, no?

soenkeliebau: The yamls in this repo are in general not linted duo to them being called .yaml.j2 - but we could do that as well, yes.

lfrancke: I'll approve and leave it up to you whether you want to change it or not

lfrancke (approved):

PR #68, opened 2022-02-10
Make Renovate slightly less annoying

teozkr:
1. Only run it during early mornings, GH bot notifications during the evening are really annoying
2. Centralize Actions PRs in operator-templating

### DIFF
--- a/playbook/update_repo.yaml
+++ b/playbook/update_repo.yaml
@@ -41,7 +41,6 @@
     - name: Find GitHub Workflow files
       find:
         paths: "{{ template_dir }}/.github/workflows/"
-        excludes: "*.j2"
         file_type: file
         hidden: true
         use_regex: false
--- a/playbook/update_repo.yaml
+++ b/playbook/update_repo.yaml
@@ -28,7 +28,7 @@
         recurse: true
       register: files_j2
 
-    - name: Find regular files that need to templating
+    - name: Find regular files that need no templating
       find:
         paths: "{{ template_dir }}/"
         excludes: "*.j2"
@@ -37,6 +37,16 @@
         use_regex: false
         recurse: true
       register: files_normal
+
+    - name: Find GitHub Workflow files
+      find:
+        paths: "{{ template_dir }}/.github/workflows/"
+        excludes: "*.j2"
+        file_type: file
+        hidden: true
+        use_regex: false
+        recurse: false
+      register: workflow_files
 
     - name: Ensure directories exist
       file:
--- a/renovate.json
+++ b/renovate.json
@@ -25,7 +25,7 @@
   "lockFileMaintenance": {
     "enabled": true,
     "schedule": [
-      "after 22:00 and before 6:00 every weekday"
+      "after 5:00 and before 6:00 every weekday"
     ]
   },
   "github-actions": {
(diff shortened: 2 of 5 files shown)


playbook/update_repo.yaml
```diff
@@ -38,6 +38,16 @@
         recurse: true
       register: files_normal
 
+    - name: Find GitHub Workflow files
+      find:
+        paths: "{{ template_dir }}/.github/workflows/"
+        excludes: "*.j2"
```
soenkeliebau, line 44: Do we really want to exclude .j2 files here? The main workflow is a .j2 file and I can see .j2 being replaced in the renovate template as well...

renovate.json
```diff
@@ -27,5 +27,8 @@
     "schedule": [
       "after 22:00 and before 6:00 every weekday"
     ]
+  },
+  "github-actions": {
+    "fileMatch": ["(^\\.github\\/workflows)\\/[^/]+\\.ya?ml$"]
```
soenkeliebau, line 32: Would this ignore .j2 workflow files?

template/renovate.json.j2
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": [
+      "after 22:00 and before 6:00 every weekday"
+    ]
+  },
+  "github-actions": {
+    // Do not update GitHub Workflows that are owned by templating
+    "ignorePaths": {[ workflow_files.files | map(attribute='path') | map('replace', '.j2', '') | map('replace', base_dir+'/template/', '') | map('regex_escape') | to_json }]
```
soenkeliebau, line 33: This renders to

` "ignorePaths": ["\\.github/workflows/daily_security\\.yml", "\\.github/workflows/reviewdog\\.yaml"]`

do we need to escape dots in there?

template/renovate.json.j2
```diff
@@ -0,0 +1,35 @@
...
+    }
+  ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": [
+      "after 22:00 and before 6:00 every weekday"
+    ]
+  },
+  "github-actions": {
+    // Do not update GitHub Workflows that are owned by templating
```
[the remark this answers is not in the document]
soenkeliebau: I'll shut up :)

renovate.json
teozkr: Oof, I thought I changed this to include `template/`.

> Would this ignore .j2 workflow files?

Yes, because it's not really practical for Renovate to modify Jinja2 templates.

template/renovate.json.j2
teozkr: Fixed

playbook/update_repo.yaml
soenkeliebau: It should actually remain valid json, as we only replace within string literals.

teozkr: Good catch, fixed.

soenkeliebau: We'll revisit this topic in a couple weeks when we did something to the template so it isn't valid json anymore and everything breaks :)

soenkeliebau (approved):

PR #79, opened 2022-02-14
Fix unneeded PRs and CRD generation error.

### LINKED ISSUE #74: List of ignored files in renovate.yaml needs to be sorted
The generated list of files that should be ignored by renovate apparantly can teigger unneeded PRs.

We might want to consider sorting these.

 `diff --git a/renovate.json b/renovate.json
 index 38a75b0..9bcf7bb 100644
 --- a/renovate.json
 +++ b/renovate.json
 @@ -28,5 +28,5 @@
        "after 5:00 and before 6:00 every weekday"
      ]
    },
 -  "ignorePaths": [".github/workflows/build.yml", ".github/workflows/reviewdog.yaml", ".github/workflows/daily_security.yml"]
 +  "ignorePaths": [".github/workflows/daily_security.yml", ".github/workflows/build.yml", ".github/workflows/reviewdog.yaml"]
  }

`

### LINKED ISSUE #75: generate-crds step is not run always
The `generate-crds` step calls `cargo build` but that actually only runs the build.rs script when said script changed.
So, most of the time it won't actually run causing weird issues with empty charts.

We 'fixed' it by introducing a `touch` before the `cargo build` but there may be better solutions

https://stackoverflow.com/questions/49077147/how-can-i-force-build-rs-to-run-again-without-cleaning-my-whole-project

soenkeliebau:
Adds a sort step to generatig the list of github action files to be ignored by renovate in the downstream repos.
fixes #74

Adds updating of timestamp on build.rs to makefile when generating crds.
fixes #75

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -63,6 +63,7 @@ clean-crds:
 	rm -rf deploy/crd/*
 
 generate-crds:
+	touch rust/operator-binary/build.rs
 	cargo build
 
 regenerate-charts: clean-crds chart-clean clean-manifests generate-crds compile-chart generate-manifests
--- a/template/renovate.json.j2
+++ b/template/renovate.json.j2
@@ -28,5 +28,5 @@
       "after 5:00 and before 6:00 every weekday"
     ]
   },
-  "ignorePaths": {[ workflow_files.files | map(attribute='path') | map('replace', '.j2', '') | map('replace', base_dir+'/template/', '') | to_json }]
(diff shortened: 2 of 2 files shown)


lfrancke (approved):
The sorting thing sounds weird but I also don't see any harm in it

soenkeliebau:
> The sorting thing sounds weird but I also don't see any harm in it

So far it has only occurred when templating was run locally by Sigi, presumably due to a different shell or something similar.
Since local runs are an important test tool though to predict whether PRs will be created or not I think it is worthwhile taking the extra precaution.

PR #89, opened 2022-03-01
- added opa bundle builder (makefile) and opa OPA_BUILDER_CLUSTERROLE…

maltesander:
… (deployment)

- removed obsolete extra_crates
- removed spaces from readme


template/Makefile.j2
```diff
@@ -15,6 +15,9 @@ VERSION := $(shell cargo metadata --format-version 1 | jq '.packages[] | select(
 ## Docker related targets
 docker-build:
 	docker build --force-rm -t "docker.stackable.tech/stackable/{[ operator.product_string }]-operator:${VERSION}" -f docker/Dockerfile .
+{[% if operator.product_string in ['opa'] %}]
```
soenkeliebau, line 18: This is probably ok for now, but if we start adding more of these conditions here I think I would prefer adding a "dockerfiles" variable that defaults to `["docker/Dockerfile"]` but can be overridden to allow building and pushing multiple docker images here.

template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2
```diff
@@ -44,6 +44,11 @@ spec:
           env:
             - name: KAFKA_BROKER_CLUSTERROLE
               value: {{ .Release.Name }}-kafka-broker-clusterrole
+{[% endif %}]
+{[% if operator.product_string in ['opa'] %}]
+          env:
+            - name: OPA_BUILDER_CLUSTERROLE
+              value: {{ .Release.Name }}-opa-builder-clusterrole
```
Comment (soenkeliebau) [inline template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2:51]: nit: should this not be called bundle-builder ?

soenkeliebau (approved):
two comments, but overall looks good to me

template/Makefile.j2
```diff
@@ -15,6 +15,9 @@ VERSION := $(shell cargo metadata --format-version 1 | jq '.packages[] | select(
...
+	docker build --force-rm -t "docker.stackable.tech/stackable/{[ operator.product_string }]-bundle-builder:${VERSION}" -f docker/Dockerfile.builder .
+{[% endif %}]
 
 docker-build-latest: docker-build
 	docker tag "docker.stackable.tech/stackable/{[ operator.product_string }]-operator:${VERSION}" \
```
soenkeliebau, line 23: Does this also need changing for the bundle builder?

soenkeliebau (changes requested):
Actually have one question that I'd like to clarify before merging..

template/Makefile.j2
razvan: We can it it's not needed. The image for the builder is never installed from `latest` but always from the exact same version as the operator.
To put it differently: the builder is image is a "product image" that always comes with the same version as the operator.

template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2
Comment (razvan) [reply template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2:51]: Yeah it should.

template/Makefile.j2
soenkeliebau: ok

soenkeliebau (dismissed):

template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2
Comment (razvan) [reply template/deploy/helm/[[product]]-operator/templates/deployment.yaml.j2:51]: Done.

razvan (approved):
lgtm

PR #94, opened 2022-03-16
Cleanup after install using microdnf

backstreetkiwi:
In our Dockerfile(s) we use microdnf but do not use the command `microdnf clean all` at the end as recommended.

With the [release 2.9.1 of hadolint](https://github.com/hadolint/hadolint/releases/tag/v2.9.1), this is added as a check and reported. (after fixing issue https://github.com/hadolint/hadolint/issues/742)

This PR adds the cleanup as recommended.


soenkeliebau:
We are actually cleaning up after installing things by rm'ing the yum cache - but that's not what the linter looks for, so it complains.

I have opened https://github.com/hadolint/hadolint/issues/796 for this and would wait to see if there is any reaction on it, before changing stuff on our end.

soenkeliebau (approved):

PR #95, opened 2022-03-17
Fix to work with yq from snap

sbernauer:
This way it works with yq from snap.
On the other hand we now have a dependency on sponge. We can resolve it by moving around tmp files if we prefer

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -42,7 +42,7 @@ chart-clean:
 	rm -rf deploy/helm/{[ operator.name }]/crds
 
 version:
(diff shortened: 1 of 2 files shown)


teozkr:
Hm, with my non-snap yq I'm able to run `yq eval -i '.asdf = "bar"' /dev/stdin < asdf.yaml`. Can you check whether that works for you too?

That should bypass the sandbox without needing `sponge`, since your shell (or `make`) would be opening the file and just passing the fd to yq.

sbernauer:
Fancy! Works fine

teozkr (approved):

PR #98, opened 2022-03-18
Update Python formatting for cargo_version.py

razvan:
... to make flake8 happy.

### DIFF
--- a/template/python/cargo_version.py
+++ b/template/python/cargo_version.py
@@ -3,9 +3,7 @@
 # Utility for viewing and managing versions of cargo workspaces and crates.
 # For workspaces, it assumes that all crate members use a single shared version.
 #
-# usage:
-# cargo_version.py [-h] [-p PROJECT] [-r] [-n {major,minor,patch}]
-#                  [-s SET] [-o] [-m PRERELEASE]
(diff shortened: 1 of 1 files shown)


template/python/cargo_version.py
```diff
@@ -21,176 +19,188 @@
 #   -m PRERELEASE, --prerelease PRERELEASE
 #                         Set pre-prelease string.
 #
-
-import argparse
-import semver
 import toml
+import semver
```
soenkeliebau, line 23: Did the linter complain about this? I still remember changing this to be alphabetic because it complained about that a few weeks ago..

razvan: I used `black` for auto-formatting and then I ran `flake8` and got no errors.

soenkeliebau: Ok. As long as the linters are happy I don't have a strong opinion either way.

Is this in line with the master version of this script in stackable/stackable-utils though?

soenkeliebau (approved):

template/python/cargo_version.py
razvan: yes

PR #100, opened 2022-03-21
Fix some variable references that lead to duplicate -operator strings in the secret operator

soenkeliebau:


sbernauer (approved):
Nice!
Running into the exact same problems with the new commons-operator

PR #115, opened 2022-04-05
Ignore values.yaml for secret-operator

teozkr:
This is required to avoid having templating revert https://github.com/stackabletech/secret-operator/pull/103. That PR is not suitable to roll out for all operators, since most operators should not need that level of privilege.


soenkeliebau (approved):
lgtm - ran it locally against two operators and confirmed that the values file in the secret operator was not updated

soenkeliebau (approved):

PR #116, opened 2022-04-14
Update action-yamlling and markdownlint

sbernauer:
Problem is described in https://github.com/reviewdog/reviewdog/issues/1158


soenkeliebau (approved):
Compared hashes, looks good to me.

PR #121, opened 2022-04-25
JQ is a requirement for test scenario templating

backstreetkiwi:
While adding the capability to run the templated integration tests in Jenkins CI, I noticed that JQ is also a requirement for the tests to run.


razvan (approved):
correct

soenkeliebau (approved):

PR #135, opened 2022-05-09
run_tests.sh: Pass on args to kuttl.

razvan:


template/scripts/run_tests.sh
```diff
@@ -25,5 +25,5 @@ popd
 
 # Run tests
 pushd tests/_work
-kubectl kuttl test
+kubectl kuttl test ${@}
```
teozkr, line 28:
```suggestion
kubectl kuttl test "$@"
```

teozkr (approved):

template/scripts/run_tests.sh
sbernauer, line 28: Hehe :P
```suggestion
kubectl kuttl test "${@}"
```

teozkr (dismissed):

soenkeliebau (approved):

PR #136, opened 2022-05-10
Add rustfmt.toml to configure formatter

sbernauer:
Merge this after https://github.com/stackabletech/secret-operator/pull/125 get's merged and didn't case any problems
I had a quick chat with @soenkeliebau and we agreed to give it a try

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,4 @@
+# This file includes unstable features, so you need to run "cargo +nightly fmt" to format your code.
+# It's also ok to use the stable toolchain by simple running "cargo fmt", but using the nigthly formatter is prefered.
+
+imports_granularity = "Crate"


soenkeliebau:
The "issue" with this is, that it requires the unstable toolchain to work, so we won't introduce any checks around this, the file is purely there as an offer, if you want to use it in your operator then that is fine and this way we ensure that is is configured the same everywhere, but if you don't feel like switching to the nightly chain to run this then that is fine too.

soenkeliebau (approved):

teozkr:
For what it's worth, you can run a single cargo command using the nightly toolchain by using `cargo +nightly fmt`.

Rust-Analyzer should also allow you to set rustfmt-specific flags (for Emacs, put `(setq rust-rustfmt-switches '("+nightly" "--edition" "2018"))` into your `.emacs`), and I assume Rust-IntelliJ has something similar.

sbernauer:
Let's try that out

Issue #140, opened 2022-05-18
Set Antora docs version automatically


fhennig:
in the operator repositories in the `docs/antora.yml` file, the `version` needs to be set.

We want to set this automatically based on the version in `Cargo.toml`.

When we are on a nightly version, the docs version should be `nightly`. When we are on a stable version, I.e. 0.10.0, the docs version should show 0.10. This is because operator features should be consistent within one minor version.

PR #141, opened 2022-05-18
Add operator version to Antora manifest

teozkr:
Doing this manually got old quickly... :P


fhennig:
Hey, thanks for picking this up! We actually came up with a ticket for it yesterday as well :D 

https://github.com/stackabletech/operator-templating/issues/140

Can you get these formatting changes in? (just 'nighly' for nightly versions and just `major.minor` for stable versions)

teozkr:
I can change it to go back to nightly for the nightlies if required, but I think for releases that seems more like a presentation problem?

fhennig:
I don' think this is just a representation issue. We talked about this yesterday, and using semantic versioning, a patch release should not contain new features or the like, so the documented featureset is identical across all patch versions. That is why we only distinguish `major.minor`. and not at the patch level.

Antora also supports a `display_version` field but that doesn't affect page IDs. if we were to just change the docs version tag to a new patch level if we release a new patch version, old links would break. This would not happen if we didn't include the patch level in the antora version.

teozkr:
Ok, done.

fhennig:
Thank you.

I ran it locally and got an error:
    
    yq eval -i '.version = nightly' /dev/stdin < docs/antora.yml
    Error: parsing expression: Lexer error: could not match text starting at 1:12 failing at 1:14.
	    unmatched text: "ni"
    make: *** [Makefile:45: version] Error 1

I ran the `test.sh` script.

This is the yq version I use: `yq (https://github.com/mikefarah/yq/) version 4.24.2`

fhennig:
While we're at it, can we set the `prerelease` key to true for the nightly version?

https://docs.antora.org/antora/latest/component-prerelease/#true

This will allow us to properly mark the latest stable version as latest in the antora playbook

fhennig (approved):
Tested, very nice, thank you very much!

PR #146, opened 2022-06-08
Remmove pylint from pre-commit.

razvan:
Apparently GH actions do not run pylint, because if it did it would have to fail.


maltesander (approved):
LGTM.

PR #149, opened 2022-06-22
Removed spark standalone from the repositories list

fhennig:
As per this ticket https://github.com/stackabletech/issues/issues/227


razvan (approved):
lgtm

PR #163, opened 2022-08-09
Added docs_templating script

fhennig:


fhennig:
in regards to the shellcheck: I can't use the bash syntax, because I can't *only* replace something at the end, only replace every occurence in the variable and I don't want that.

razvan (dismissed):
lgtm

razvan (approved):

PR #164, opened 2022-08-10
Fix bash linter warnings from #163

sbernauer:
As this creates failing PRs in the downstream repos e.g. https://github.com/stackabletech/druid-operator/pull/283

### DIFF
--- a/template/scripts/docs_templating.sh
+++ b/template/scripts/docs_templating.sh
@@ -33,7 +33,7 @@ fi
 find "$docs_dir" -name '*.j2' |
 while read -r file
 do
-  new_file_name=${file//.j2/}
+  new_file_name=${file%.j2} # Remove .j2 suffix
   echo "templating $new_file_name"
   jinja2 "$file" "$templating_vars_file" -o "$new_file_name"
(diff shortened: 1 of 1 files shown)


template/scripts/docs_templating.sh
```diff
@@ -30,9 +30,10 @@ then
   echo "$templating_vars_file does not exist, cannot start templating."
 fi
 
-for file in $(find "$docs_dir" -name '*.j2')
+find "$docs_dir" -name '*.j2' |
+while read -r file
 do
-  new_file_name=$(echo "$file" | sed 's/.j2$//')  # cut of the '.j2'
+  new_file_name=${file//.j2/}
```
fhennig, line 36: This solution was discussed, but it isn't identical to the `sed` variant. the new version will replace *all* .j2 occurrences. I ignored the linter in that case, because it was only a suggestion.

sbernauer: Correct! Updated to only remove the suffix

fhennig (approved):
thanks! LGTM

PR #166, opened 2022-08-11
Cross platform operator compiling

Maleware:
Made multi-arch builds possible.

In Makefile.j2:
Added docker-build-mulit-arch as an additional possibility to use make, multi-arch images will be build there. 

In build.yml.j2:
A new job for build, will be executed in addition to single arch images.

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -253,7 +253,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up Helm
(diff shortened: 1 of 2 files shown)


razvan (commented):
You can use the `./test.sh` script in this repo to test the `Makefile` generation locally. To make it faster, edit `repositories.yaml` to only test on one repo.

lfrancke:
This is in the refinement column now but it is already a pull request.
Github unfortunately doesn't show those project events yet.
Can anyone help me figure out what this is about?

PR #170, opened 2022-08-23
Configure yamllint to accept non-indented sequences

siegfriedweber:
Configure yamllint to accept non-indented sequences

`serde_yaml:0.9` does not indent sequences anymore, so yamllint should not complain about it.


maltesander (approved):
LGTM.

PR #171, opened 2022-08-24
Don't assume that bash is at /bin/bash

teozkr:
This broke the build on NixOS, which _only_ provides /bin/sh and /usr/bin/env, and expects everything else to be found on the $PATH.


siegfriedweber (approved):
Thank you! 👍

PR #176, opened 2022-09-06
Use GitHub CLI package from Stackable Nexus

### LINKED ISSUE #174: Templating Action is broken due to expired gpg key of github cli
Templating is currently not working as the step to install the GH cli in the container [fails](https://github.com/stackabletech/operator-templating/runs/8182091013?check_suite_focus=true).

This is due to https://github.com/cli/cli/issues/6175

A few options are mentioned in the comments for that issue, but at this moment there is no fix available.

Options are: 
1. Install unauthenticated package, but this gets credentials for our github account and makes changes to repos - the security impact if compromised might be severe
2. Download binary from gh releases page instead of using apt 

From browsing the issue it [seems](https://github.com/cli/cli/issues/6175#issuecomment-1235984381) like updating the expired key may not be happening anytime soon, if ever. Item 2 from the list above is the recommended solution for now and might turn into the only supported one soon.

I think we should go with option 2 to fix this.

siegfriedweber:
Closes #174

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -26,10 +26,9 @@ jobs:
 
       - name: Install Github CLI
         run: |
-            curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg && \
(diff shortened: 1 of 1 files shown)


teozkr (commented):
I agree that we don't want this in deb-release, but would it be worth setting up a deb-tools repo or something instead of downloading the debs by hand?

siegfriedweber:
For the tools we use, it is an exception that a Debian package is provided and that this Debian package is not already contained in a proper repository. We could also use the `gh` binary provided in the tar.gz file. Then we would do it similar to `protoc`. Both tools can be found side by side in the packages directory of Nexus, not one in packages and the other one in deb-tools.

I vote for keeping `gh` in packages but I am open to change this.

teozkr (approved):

PR #179, opened 2022-09-08
Install protoc in GitHub workflow

### LINKED ISSUE #178: Secret operator breaks the templating playbook runs
The templating playbooks run 'generate manifests' to check if manifests are up to date. For the secret operator this seems to require a protoc binary to be present, which currently it isn't and causes the run to fail with

> Could not find `protoc` installation and this build crate cannot proceed without

as can be seen here: https://github.com/stackabletech/operator-templating/runs/8242005122?check_suite_focus=true

This is done when:
- [ ] the action completes again without errors

siegfriedweber:
Fixes #178 

successful run: https://github.com/stackabletech/operator-templating/actions/runs/3013408792

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -30,6 +30,10 @@ jobs:
             sudo apt-get install /opt/gh.deb && \
             sudo rm /opt/gh.deb
 
+      - name: Install protoc
+        run: |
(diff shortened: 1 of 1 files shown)


sbernauer (approved):
Nice thx!

PR #182, opened 2022-09-19
Pin rustc version

teozkr:
See https://github.com/stackabletech/issues/issues/244

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -62,6 +62,10 @@ repositories:
     product_string: zookeeper
     url: stackabletech/zookeeper-operator.git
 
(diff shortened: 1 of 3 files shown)


sbernauer (approved):
We have problems with 1.64.0, don't we? So 1.63.0 LGTM

teozkr:
1.64.0 isn't out yet, for one…

sbernauer:
Oh yeah ^^ I just looked at https://github.com/rust-lang/rust/blob/master/RELEASES.md and didn't noticed it's dated in the future :see_no_evil:

PR #183, opened 2022-09-19
Enable renovate to update the ubi8 version

Maleware:
Pinning ubi8-minimal to the latest version [8.6-941](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=overview&push_date=1663055915000&architecture=amd64)


template/docker/Dockerfile.j2
```diff
@@ -4,7 +4,7 @@
 # =============
 FROM docker.stackable.tech/stackable/ubi8-rust-builder AS builder
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal AS operator
+FROM registry.access.redhat.com/ubi8/ubi-minimal@sha256:37b1bbdc042d32573738c72e4e0f0592b4198a5de1c3a3f32515d29ba4605488 AS operator
```
soenkeliebau, line 7: If we pin the version like this Renovate won't track the tag with the version number in the string, which I think is helpful.

If we commit it like this:
> FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6 AS operator

then renovate should probably pin it as 

> FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6@sha256:e58664de16551db29fb0eaaeb3c4a44eaf95ad89a5b2399a1107041c4f2d6d34 as operator

While still updating versions.

PR #193, opened 2022-10-18
Post-README tweaks

vsupalov:
* Sync templating with desired state of operator repositories.
* Add tweaks to fix issues with rendered READMEs


lfrancke (approved):
I haven't tested it but the canges look reasonable to me
I can't see where no_jenkins_job_badge comes from but I assume that's elsewhere?

vsupalov:
Thanks! Yes, the variable will be set in single operators and is not further used here.

PR #197, opened 2022-10-25
Fix incorrect README docs links

vsupalov:
The operator doc paths differ from the repository names, and require a separate variable, set per operator.


fhennig (approved):
LGTM!

The inconsistencies here are unfortunate

PR #209, opened 2022-11-23
Bump rustc to 1.65.0

teozkr:


teozkr:
Isn't that what we have CI for? ;)

PR #212, opened 2022-11-24
Add secondary Nix-based build system

teozkr:
This won't be used for prod builds (for now), but allows Nix to cache library builds, which gives us _much_ faster feedback loops during development (for operators that need to run inside of the K8s cluster).

This is based on (and obsoletes) the Nix build system from secret-op and listener-op.


teozkr:
I have tested a bit against darwin/macOS and the basic principles still work (using a remote Linux builder). However, there are some surrounding papercuts that need to be resolved for that setup to be usable (for example, we need to configure the target platform, `streamLayeredImage` produces a Linux-only script that won't run on darwin, and `crate2nix` doesn't build native darwin binaries correctly because of some dependency issues).

template/default.nix
```diff
@@ -29,6 +29,7 @@ rec {
     tag = dockerTag;
     contents = [ pkgs.bashInteractive pkgs.coreutils pkgs.util-linuxMinimal ];
     config = {
+      Env = ["PRODUCT_CONFIG=${deploy/config-spec/properties.yaml}"];
```
nightkr, line 32: Hm, we might need to think a bit deeper about this before adding it, not all of our operators have product-configs at all.

template/Makefile.j2
```diff
@@ -73,3 +73,6 @@ regenerate-charts: chart-clean compile-chart
 build: regenerate-charts helm-package docker-build
 
 publish: build docker-publish helm-publish
+
+run-dev:
+	nix run -f. tilt -- up --port 543{[operator_index}]
```
nightkr, line 78: This is going to go from `5439` to `54310`, which is pretty surprising. Maybe `--port {[5430 + operator_index}]` would be a better fit?

template/default.nix
nightkr: One way would be to make it conditional on properties.yaml existing:

```nix
Env = lib.optional (builtins.pathExists deploy/config-spec/properties.yaml) "PRODUCT_CONFIG=${deploy/config-spec/properties.yaml}";
```

template/Makefile.j2
soenkeliebau: It would indeed, didn't know that was possible, which is why I was willing to live with that jump :)

template/default.nix
soenkeliebau: done

soenkeliebau:
addressed comments @nightkr

nightkr:
LGTM now. I can't approve since it was my PR originally.

soenkeliebau (approved):

soenkeliebau:
I can :)

nightkr:
Want the honours of merging/starting the rollout? :P

soenkeliebau:
I only did some drudge work here, its your brainchild :)

PR #213, opened 2022-12-15
Release branch prep 001

razvan:
For this repo:

* Some Ansible task improvements.

For each operator (secret, commons, zookeeper):

* New .github/workflow/build.yaml
* New Makefile


template/.github/workflows/build.yml.j2
```diff
@@ -9,7 +9,7 @@ on:
       - trying
       - "renovate/**"
     tags:
-      - "*"
+      - '[0-9][0-9].[0-9]+.[0-9]+'
```
soenkeliebau, line 12: Is it on purpose that we require a two digit major number? Or am I misunderstanding the regex?
Also, does the . need to be escaped to match an actual . ?

adwk67: Well, we're at 22.11 already :) That's not a regex but a github-action specific glob pattern, for which there are only a limited number of special characters (otherwise we could have limited the second part to a number 1-12, for the month).

soenkeliebau (approved):

PR #217, opened 2022-12-23
Fix Ansible warnings

razvan:
Fixes the following warning when generating kuttl tests:
```
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'
[DEPRECATION WARNING]: "include" is deprecated, use include_tasks/import_tasks instead. This feature will be removed in version 2.16. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
```

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1,50 +1,50 @@
 ---
 repositories:
-#  - name: airflow-operator
-#    pretty_string: Apache Airflow
-#    product_string: airflow
-#    url: stackabletech/airflow-operator.git
+  - name: airflow-operator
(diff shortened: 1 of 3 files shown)


lfrancke (approved):
Looks good to me but I don't see where you removed any `include` so not sure how that warning is tackled by this PR

lfrancke:
@razvan I just stumbled across this old PR...

razvan:
closing since we don't use ansible for tests anymore.

PR #223, opened 2023-01-31
Migrated test expansion from Ansible to beku.py

backstreetkiwi:
The expansion of the Kuttl templates is done by beku.py now. 

(see https://github.com/stackabletech/issues/issues/315 for details)


fhennig (approved):
tested locally, LGTM!

PR #232, opened 2023-02-08
Fix most deprecation warnings in our builds

lfrancke:
Together with the latest round of Renovate updates most of the errors mentioned below should be fixed.
There is still audit-check and clippy-check for which we need an alternative but I wanted something out of the door and work on the others later.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af, actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505. For more information see: 
```

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```


template/.github/workflows/build.yml.j2
```diff
@@ -38,21 +38,14 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
         with:
           submodules: recursive
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # renovate: tag=v1.0.7
-        with:
-          profile: minimal
-          toolchain: "{[ rust_version }]"
-          override: true
+      - uses: dtolnay/rust-toolchain@{{ rust_version }}
```
fhennig, line 41: Does this work? The curly braces were with brackets before

lfrancke: Well, that is a good question..... I thought this was ansible/Jinja...

Good catch. Paging @soenkeliebau

soenkeliebau: We [changed](https://github.com/stackabletech/operator-templating/blob/main/playbook/update_repo.yaml#L68) the variable delimiting pattern for historic reasons. We used to apply this to rpmspec files, which also use {{ }} to delimit stuff and confused Ansible.

So correct is `{[ rust_version }]`

soenkeliebau:
I'm about to merge a bit of stuff into the operators anyway, if this is ready to go I could include it?

lfrancke:
Yes please, I've just been waiting for a review.

soenkeliebau:
I'm afraid I cannot really comment on most of the changes, would require a bit of research on my part ..

@fhennig  you reviewed this before, can you take another look?

fhennig:
I didn't really review it, I just had a glance and noticed something. there is still a yamllint error, and on the same line (41) it looks like the indentation isn't correct.

Other than that it _looks_ fine to me, and I can't really test it locally or anthing, so I could stamp my approval on it :laughing:

lfrancke:
I'll fix the YAML lint, sorry, I thought I already did
If we don't get this in all our builds will start failing at the end of April so I'd like to not let this PR rot too long

lfrancke:
@soenkeliebau @fhennig Ready to "review" again. I suggest we just try merging this and I will only merge a single PR (e.g. NiFi) to test

fhennig (approved):

lfrancke:
Well....most deprecation warnings are gone (I know about the remaining ones) but there were at least two mistakes in there anyway so I'll clean up the PRs and will create a follow-up

PR #233, opened 2023-02-13
Update Tiltfile to enable users overriding the default registry with custom values

soenkeliebau:
Currently we hardcode the registry to be used for pushing docker images to stackables sandbox registry.
While this is fine for Stackable members, it won't work for external contributors. This PR adds the ability to create a config file for Tilt in which different registries can be specified that will then be picked up.

Since this file should only be used for user specific settings, it is also added to the .gitignore file to avoid it accidentally ending up in vc.

### DIFF
--- a/template/Tiltfile
+++ b/template/Tiltfile
@@ -1,5 +1,8 @@
+# If tilt_options.json exists read it and load the default_registry value from it
 settings = read_json('tilt_options.json', default={})
 registry = settings.get('default_registry', 'docker.stackable.tech/sandbox')
+
+# Configure default registry either read from config file above, or with default value of "docker.stackable.tech/sandbox"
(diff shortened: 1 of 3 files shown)


template/Tiltfile
```diff
@@ -22,7 +27,7 @@ helm_crds, helm_non_crds = filter_yaml(
       'deploy/helm/' + operator_name,
       name=operator_name,
       set=[
-         'image.repository=docker.stackable.tech/sandbox/' + operator_name,
+         registry + '/' + operator_name,
```
nightkr, line 30: You still need the `image.repository=`.

soenkeliebau: Ups

nightkr (approved):

PR #238, opened 2023-02-15
Fix cargo install (really for real maybe)

lfrancke:
This is a bit embarrasing...fourth time lucky...


sbernauer (approved):
Not tested, see you in the next fixup PR

PR #239, opened 2023-02-21
Update Review Checklist

lfrancke:


lfrancke:
Good points.
Tasklists work in PRs, yes. I tried: https://github.com/stackabletech/kafka-operator/pull/557

Good point about the audience....

I could split the list in three parts but some of those things can be done by either party.
Do we want a rule for this? I'd say no but I agree that there is confusion, so it might be best to come up wtih something and just stick to that until we change it.

## Reviewer
- Code contains useful comments
- (Integration-)Test cases added (or not applicable)
- Changelog updated (or not applicable)
- Documentation added or updated (or not applicable)
- Cargo.toml only contains references to git tags (not specific commits or branches)

## Author
- Changes are OpenShift compatible
- CRD change approved (or not applicable)
- Helm chart can be installed and deployed operator works (or not applicable)

## Acceptance
- Feature Tracker has been updated (or not applicable)
- Proper release label has been added (or not applicable)

fhennig:
Yes I think _someone_ should be made to tick the boxes, if it's not specified, boxes will remain unticked. The split looks good to me, but I'd put "Author" first because they are chronologically first I suppose

lfrancke:
Will do, thanks! I will update and ask for feedback once more and then we can change again as needed.

template/.github/pull_request_template.md
```diff
@@ -5,14 +5,31 @@
...
+## Definition of Done Checklist
 
+- Not all of these items are applicable to all PRs, the author should update this template to only leave the boxes in that are relevant
+- Please make sure all these things are done and tick the boxes
+ 
+```[tasklist]
+# Author
+- [ ] Changes are OpenShift compatible
+- [ ] CRD changes approved
+- [ ] Helm chart can be installed and deployed operator works
```
sbernauer, line 17: What about `For non-trivial PRs: Integration tests passed`?

lfrancke: added (for the author) vs passed (for the reviewer) you mean?

sbernauer: I meant passed. Normally the author starts the tests, so I'm fine with him checking the mark when the tests pass

lfrancke: Just so I get it right: We already have "Integation tests added" for the reviewer.
Do you suggest to _move_ that to _Author_ and rename it to "passed" instead of added or do you suggest to have both?

sbernauer: I would go with both

soenkeliebau (approved):

PR #242, opened 2023-03-07
Add Kerberos libraries to the Docker images

nightkr:
This is required for https://github.com/stackabletech/secret-operator/pull/99. At some point we'd need to add extension hooks to let each operator decide some of these things, but... not yet.

### DIFF
--- a/template/docker/Dockerfile.j2
+++ b/template/docker/Dockerfile.j2
@@ -25,7 +25,7 @@ RUN microdnf install -y yum \
 
 COPY LICENSE /licenses/LICENSE
 
-COPY --from=builder /app/stackable-{[ operator.name }] /
+COPY --from=builder /app/* /usr/local/bin
 {[% if operator.include_productconfig is undefined or operator.include_productconfig == true %}]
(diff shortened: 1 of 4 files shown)


template/docker/Dockerfile.j2
```diff
@@ -23,9 +23,12 @@ RUN microdnf install -y yum \
   && yum clean all \
   && microdnf clean all
 
+# Install kerberos client libraries
+RUN microdnf install -y krb5-libs libkadm5 && microdnf clean all
+
 COPY LICENSE /licenses/LICENSE
 
-COPY --from=builder /app/stackable-{[ operator.name }] /
+COPY --from=builder /app/* /usr/local/bin
```
nightkr, line 31: This is technically breaking, since anyone assuming that the operator would be available in / will be broken... but as long as you relied in the ENTRYPOINT you should be fine.

soenkeliebau: Fine for me in principle, but I have to admit I don't fully understand where the need for this change comes from?

razvan (approved):
lgtm

template/docker/Dockerfile.j2
soenkeliebau: Ah, thanks!

PR #245, opened 2023-04-17
Cross-platform fix

Maleware:
This fixes that MacOS can't use dev/stdin to fetch input from. 

Instead we moved to cat into a new file then mv it to the original one


sbernauer (approved):
Works for me

PR #260, opened 2023-06-22
Feature/multi arch operator

Maleware:
This PR enables multiarch builds for operator. 

Changes are made to:

- Dockerfile.j2 : Pointing to the specific manifest list
- Makefile.j2 : Introducing a variable to pass a platform annotation and a new mode manifest-push to create a manifest list
-

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -9,6 +9,7 @@
 
 .PHONY: build publish
 
+ARCH := ""
 TAG    := $(shell git rev-parse --short HEAD)
 OPERATOR_NAME := {[ operator.name }]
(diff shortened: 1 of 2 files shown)


template/Makefile.j2
```diff
@@ -9,6 +9,7 @@
 
 .PHONY: build publish
 
+ARCH :=
```
razvan, line 12: What is the plan with this in the future ?

razvan (commented):
You need to update the `print-docker-tag` target too. It's used in the GH workflow `build.yaml `

razvan (commented):
You need to update the `print-docker-tag` target too. It's used in the GH workflow `build.yaml `

PR #262, opened 2023-06-28
Apply default resource limits to operator containers

soenkeliebau:
The following resource limits will be applied to all operators by default:
resources:
  limits:
    cpu: 100m
    memory: 128Mi
  requests:
    cpu: 100m
    memory: 128Mi

These will only be applied if users do not specify resources in their values.yaml file. Anything specified there will override the defaults.


soenkeliebau:
> Haven't looked into the code: Can we give commons-op a bit higher resources (e.g. 200m and 1024Mi) as it does watch all the Pods, and customers had problems in the past?

Easiest way would be to put that into the values.yaml of commons..

Failing that ... I could introduce variables that get written out of course which would allow us to override this per operator.

soenkeliebau:
ping @sbernauer 
done

sbernauer:
Thx! Currently don't have the time to do a proper review of all the helm stuff, maybe someone with better helm knowledge is better suited. I would try to finish the resources on products first if that is ok?

repositories.yaml
```diff
@@ -9,6 +32,12 @@ repositories:
     pretty_string: Stackable Commons
     product_string: commons
     url: stackabletech/commons-operator.git
+    resources:
+      cpu_limit: 200m
+      memory_limit: 1024Gi
+      cpu_request: 200m
+      memory_request: 1024Gi
```
nightkr, line 39: Not a huge fan of maintaining these overrides _here_, but I can see why it's done that way…

template/deploy/helm/[[operator]]/resources/default_node_resources.yaml.j2
```diff
@@ -0,0 +1,6 @@
+limits:
+  cpu: {[ operator.daemonset_resources.cpu_limit  | default(resource_defaults.cpu_limit) }]
```
Comment (nightkr) [inline template/deploy/helm/[[operator]]/resources/default_node_resources.yaml.j2:2]: We're pretty inconsistent about referring to this as `node` or `daemonset` right now, we should at least pick one.

template/deploy/helm/[[operator]]/templates/_helpers.tpl
```diff
@@ -74,3 +74,46 @@ Labels for Kubernetes objects created by helm test
...
+{{- define "operator.resources" -}}
+{{- if .Values.resources }}
+{{- .Values.resources | toYaml }}
+{{- else }}
+{{- .Files.Get "resources/default_resources.yaml" }}
+{{- end }}
+{{- end }}
+
+{{/*
+Similar to operator.resources, but meant to be used for setting resources on daemonsets.
```
Comment (nightkr) [inline template/deploy/helm/[[operator]]/templates/_helpers.tpl:95]: This comment is for `.node`, but it's on the `.controller`

template/deploy/helm/[[operator]]/templates/_helpers.tpl
```diff
@@ -74,3 +74,46 @@ Labels for Kubernetes objects created by helm test
...
+As some operators have both, we wanted to be able to define different values.
+See comment on `operator.resources` for more details.
+*/}}
+{{- define "operator.controller.resources" -}}
```
Comment (nightkr) [inline template/deploy/helm/[[operator]]/templates/_helpers.tpl:99]: IMO, we should pick either this or the "bare" `.resources` across the whole platform.

Comment (soenkeliebau) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:99]: in that case I'd vote for keeping it the bare resources to avoid "cosmetic" (but strictly speaking breaking) changes througout all operators.

repositories.yaml
soenkeliebau: Me neither, but I thought it was preferrable to doing it inline in the templates and couldn't come up with a better way.

nightkr: Actually, can't the operators put it in their values.yaml? That way it would override the templated values but still be overridden in turn by any user-provided config.

template/deploy/helm/[[operator]]/templates/_helpers.tpl
Comment (nightkr) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:99]: I'm fine with that (arguably the controllers are the "primary" components), but don't we end up doing this anyway since both variants are new?

repositories.yaml
soenkeliebau: They sort of already do:

https://github.com/stackabletech/nifi-operator/blob/8f1797e1137207547bdcbddc38c122e48b50c0e2/deploy/helm/nifi-operator/values.yaml#L34

My intention was to keep this as is, to avoid changing everything but have a default in place that is enforced if the user doesn't set anything.

But maybe I misunderstood your intention.

template/deploy/helm/[[operator]]/templates/_helpers.tpl
Comment (soenkeliebau) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:99]: `.resources` exists already and is picked up by the helm chart if set.

https://github.com/stackabletech/nifi-operator/blob/8f1797e1137207547bdcbddc38c122e48b50c0e2/deploy/helm/nifi-operator/values.yaml#L34

soenkeliebau:
Closed this in favor of directly setting resource limits in the values.yaml of the operators to limit complexity.

PR #268, opened 2023-07-12
Sign container images in Nexus

dervoeti:
Needed for https://github.com/stackabletech/issues/issues/370

Enables signing of our operator images (we will need to sign our product images too), publishes signatures in Nexus.
We plan on migrating our artifacts to Harbor later, which will allow us to sign Helm charts, too.

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -30,16 +30,16 @@ docker-build:
 
 docker-publish:
 	echo "${NEXUS_PASSWORD}" | docker login --username github --password-stdin "${DOCKER_REPO}"
-	docker push --all-tags "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}"
-	# Obtain a reference to the pushed image by digest, because signing by tag is deprecated and will be removed from cosign in the future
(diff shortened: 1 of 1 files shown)


template/Makefile.j2
```diff
@@ -31,6 +31,12 @@ docker-build:
 docker-publish:
 	echo "${NEXUS_PASSWORD}" | docker login --username github --password-stdin "${DOCKER_REPO}"
 	docker push --all-tags "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}"
+	REPO_ARTIFACT_BY_DIGEST=$$(docker inspect --format='{{range .RepoDigests}}{{ . }}{{end}}' '${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:${VERSION}' | grep -E '^${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@sha256:[0-9a-f]{64}$$' | head -n1);\
```
lfrancke, line 34: This could use a comment on what is happening here and why and....where does the pushing happen?
Is that `cosign sign`?

dervoeti: Yes, `cosign sign` automatically pushes the signature. Added some comments :slightly_smiling_face:

dervoeti:
I simplified the way the image digest is obtained, it should be clearer now and it's also almost the same way we will do it for Helm charts (Helm just outputs the digest to stderr). Tested it again by generating the `Makefile` for airflow-operator and triggering the build, worked fine (Workflow run: https://github.com/stackabletech/airflow-operator/actions/runs/5535650288).

Techassi (approved):
LGTM!

PR #269, opened 2023-07-13
fix: Template default.nix file for secret-operator

### LINKED ISSUE #267: Update `default.nix` file to use correct Kerberos crate overrides
This change makes sure the correct crate overrides are set when building the `secret-operator`. This conflict came up when merging #264 (tracked in #263) in the downstream repos.

```[tasklist]
### Tasks
- [ ] https://github.com/stackabletech/operator-templating/pull/269
```

Techassi:
This alters the `default.nix` file for the secret operator. This file is slightly different compared to other operators due to Kerberos. Fixes #267

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -35,7 +35,6 @@ jobs:
           libssl-dev \
           pkg-config
 
-
       - name: Install jinja2
         run: pip install -r requirements.txt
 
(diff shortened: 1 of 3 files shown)


siegfriedweber (approved):

nightkr:
None of this needs to be Jinja-templated, only `secret-op` provides `krb5-sys` or `libgssapi-sys`, and they don't harm anyone if they're unused (similar logic appleis to the others: `stackable-secret-operator` is only used in.. the secret-operator, and prost/tonic are only used in the CSI operators (secret, listener)).

PR #276, opened 2023-08-01
Improve performance of builds

lfrancke:
- Things installed via apt should be cached now
- Use pre-built binary of cargo-udeps and cargo-set-version
- Update Helm to 3.13


lfrancke:
I did test this back in August and just started a new test over here https://github.com/stackabletech/nifi-operator/pull/543

NickLarsenNZ (dismissed):
LGTM

lfrancke:
Thank you!
A test revealed a mistake.
And there is the question whether we want to use the pre-built binaries or rely on the cache.
I think I prefer the pre-built ones because a) we have limited cache space and I'd rather use it for other things and b) the cache will be wiped if things change (e.g. dependencies, toolchain etc.)

lfrancke:
This also builds on top of https://github.com/stackabletech/operator-templating/pull/303 so that should be reviewed and merged first

NickLarsenNZ (approved):
LGTM, I have approved the dependee PR too

PR #280, opened 2023-09-02
Add cargo test to pre commit hooks

razvan:


template/.pre-commit-config.yaml.j2
```diff
@@ -42,3 +42,12 @@ repos:
         entry: make regenerate-charts
         stages: [commit, merge-commit, manual]
         pass_filenames: false
+
+  - repo: local
+    hooks:
```
Techassi, line 47: Lets combine it with the above repo. `hooks` accepts a list of hooks and thus `cargo-test` can be combined with `regenerate-charts` above.

Techassi (changes requested):
Minor stuff, lgtm otherwise.

template/.pre-commit-config.yaml.j2
razvan: Good idea. Done.

Techassi (approved):
LGTM!

PR #283, opened 2023-09-11
Normalize nix build

nightkr:
Hoisted `csi-grpc` rules from listener-op, and removed the conditionals from #269.

This doesn't need to be rolled out immediately, the applicable repos have already been fixed.


siegfriedweber (approved):
LGTM

PR #284, opened 2023-09-12
Upgrade nixpkgs in nix/sources.json to a revision with Rust 1.71.1

siegfriedweber:
Upgrade nixpkgs in nix/sources.json to a revision with Rust 1.71.1

Our build pipeline uses Rust 1.71.1, so the same version should be used when building with Nix. The commit 676fe5e01b9a41fa14aaa48d87685677664104b1 is the one before Rust was upgraded to version 1.72.0.

### DIFF
--- a/template/nix/sources.json
+++ b/template/nix/sources.json
@@ -17,10 +17,10 @@
         "homepage": "",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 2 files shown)


sbernauer (approved):
Thanks!

nightkr (dismissed):
niv's boilerplate has also had an update, we could use another `niv init`

PR #289, opened 2023-11-16
Update crate2nix and set CARGO_MANIFEST_LINKS

sbernauer:
Previously we were running into https://github.com/briansmith/ring/issues/1786.
We need to update to pull in https://github.com/nix-community/crate2nix/pull/314.
Also, https://github.com/NixOS/nixpkgs/pull/266787 is not merged yet, so we set the env var to the expected value manually.
Once https://github.com/NixOS/nixpkgs/pull/266787 is merged, we should remove the manual env var.


NickLarsenNZ (approved):
I saw that this worked :+1:

PR #299, opened 2023-12-12
fix: include templating step in the local test run

NickLarsenNZ:
This allow `./test.sh` to run the templating locally


razvan (approved):
lgtm

PR #304, opened 2023-12-18
Ask for operator and product versions separately for bug reports

nightkr:


template/.github/ISSUE_TEMPLATE/bug_report.yml.j2
```diff
@@ -9,8 +9,13 @@ body:
...
     attributes:
-      label: Affected version
-      description: Which version do you see this bug in?
+      label: Affected Stackable version
+      description: Which version of the Stackable Operator do you see this bug in?
+
+  - type: input
+    attributes:
+      label: Affected {[ operator.pretty_string }] version
+      description: Which version of {[ operator.pretty_string }] do you see this bug in?
```
nightkr, line 18: Ideally this should be filtered out for meta operators (commons, secret, listener), but we don't currently have that information in repositories.yaml.

fhennig (dismissed):

template/.github/ISSUE_TEMPLATE/bug_report.yml.j2
nightkr: 92017a4805d99333005c3dcec103a0da542c4118

fhennig (approved):
looks sensible!

PR #312, opened 2024-01-25
Publish SBOMs

dervoeti:
Fixes https://github.com/stackabletech/issues/issues/398.

Successful test run with airflow-operator: https://github.com/stackabletech/airflow-operator/actions/runs/7652323720/job/20851912198

The verified SBOM for the airflow-operator can be fetched with:
```
cosign verify-attestation --type cyclonedx --certificate-identity-regexp '^https://github.com/stackabletech/.+/.github/workflows/.+@.+' --certificate-oidc-issuer https://token.actions.githubusercontent.com oci.stackable.tech/sdp/airflow-operator:0.0.0-dev | jq '.payload' -r | base64 -d | jq '.predicate' -r > verified_sbom.json
```

Notes and possible next steps:
 - I also removed an incorrect colon from the image references. Before, it was `${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:@$$REPO_DIGEST_OF_IMAGE` which results in something like `oci.stackable.tech/sdp/airflow-operator:@sha256:b2707e4a282844d314aa4ee261b4680c91c0c5f484ed6d235282d7c4ff10bda2`, which is an invalid reference (but cosign ignored it and treated it correctly). Now it's `...airflow-operator@sha256:b27...` instead of `...airflow-operator:@sha256:b27...` (first colon removed).
 - I handcraft PURLs for the operator images in the `Makefile`. They look like this: `pkg:docker/sdp/airflow-operator@sha256:50c2df6011949ded5e17cf1c85facf568ffdcca679664feb540dd300ea7c07f5?repository_url=oci.stackable.tech`. I verified that one using [packageurl.rs](https://github.com/package-url/packageurl.rs):
```
PackageUrl {
    ty: "docker",
    namespace: Some(
        "sdp",
    ),
    name: "airflow-operator",
    version: Some(
        "sha256:50c2df6011949ded5e17cf1c85facf568ffdcca679664feb540dd300ea7c07f5",
    ),
    qualifiers: {
        "repository_url": "oci.stackable.tech",
    },
    subpath: None,
}
```
 - Syft adds all the labels from the container image to the SBOM under `.metadata.properties`. This includes a description of the UBI minimal image, which we could remove from the image labels during the build process.
 - Out operator images contain Python (`/usr/libexec/platform-python3.6`), we could check whether it's needed and where it comes from.
 - We need to document to our customers how to get and verify the SBOMs. Customers can enforce the presence of a valid SBOM before starting a container via tools like Kyverno. Since SBOMs are published in Harbor this will first be relevant once we pull our images from Harbor by default.
 - When I ran Syft version 0.101.1 on my Macbook, it reported all the crates for airflow-operator twice: once with our generated SBOM as source (the one generated by `cargo cyclonedx`), once with the operator binary as source. This did not happen on Linux (even though it was the same version), there it only reported the crates with the SBOM as source (which is what we want). I haven't debugged this further and just passed the `--exclude "/usr/local/bin/stackable-${OPERATOR_NAME}"` flag to Syft to make sure it does not try to analyze the binary, that way it produced the correct output for me on MacOS and Linux.

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -335,7 +335,10 @@ jobs:
         # Recreate charts and publish charts and docker image. The "-e" is needed as we want to override the
         # default value in the makefile if called from this action, but not otherwise (i.e. when called locally).
         # This is needed for the HELM_REPO variable.
-      - uses: sigstore/cosign-installer@9614fae9e5c5eddabb09f90a270fcb487c9f7149 # tag=v3.3.0
+      - name: Install cosign
+        uses: sigstore/cosign-installer@9614fae9e5c5eddabb09f90a270fcb487c9f7149 # tag=v3.3.0
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0.15.5
       - name: Publish Docker image and Helm chart
         run: make -e publish
         # Output the name of the published image to the Job output for later use
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -45,7 +45,7 @@ docker-publish:
 	fi;\
 	# This generates a signature and publishes it to the registry, next to the image\
 	# Uses the keyless signing flow with Github Actions as identity provider\
-	cosign sign -y ${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:@$$REPO_DIGEST_OF_IMAGE
+	cosign sign -y ${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE
 
 	# Push to Harbor
 	# We need to use "value" here to prevent the variable from being recursively expanded by make (username contains a dollar sign, since it's a Harbor bot)
@@ -59,7 +59,18 @@ docker-publish:
 	fi;\
 	# This generates a signature and publishes it to the registry, next to the image\
 	# Uses the keyless signing flow with Github Actions as identity provider\
-	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:@$$REPO_DIGEST_OF_IMAGE
+	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE;\
+	# Generate the SBOM for the operator, this leverages the already generated SBOM\
+	syft scan -o cyclonedx-json --exclude "/usr/local/bin/stackable-${OPERATOR_NAME}" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}" ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE > sbom.json;\
+	# Determine the PURL for the container image\
+	PURL="pkg:docker/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE?repository_url=${OCI_REGISTRY_HOSTNAME}";\
+	# Get metadata from the image\
+	IMAGE_DESCRIPTION=$$(docker inspect --format='{{.Config.Labels.description}}' ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION});\
+	IMAGE_NAME=$$(docker inspect --format='{{.Config.Labels.name}}' ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION});\
+	# Merge the SBOM with the metadata for the operator\
+	jq -s '{"metadata":{"component":{"description":"'"$$IMAGE_NAME. $$IMAGE_DESCRIPTION"'","supplier":{"name":"Stackable GmbH","url":"https://stackable.tech/"},"author":"Stackable GmbH","purl":"'"$$PURL"'","publisher":"Stackable GmbH"}}} * .[0]' sbom.json > sbom.merged.json;\
+	# Attest the SBOM to the image\
+	cosign attest -y --predicate sbom.merged.json --type cyclonedx ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE
 
 # TODO remove if not used/needed
 docker: docker-build docker-publish
@@ -85,7 +96,7 @@ helm-publish:
 	docker login --username '${value OCI_REGISTRY_SDP_CHARTS_USERNAME}' --password '${OCI_REGISTRY_SDP_CHARTS_PASSWORD}' '${OCI_REGISTRY_HOSTNAME}';\
 	# This generates a signature and publishes it to the registry, next to the chart artifact\
 	# Uses the keyless signing flow with Github Actions as identity provider\
-	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_CHARTS}/${HELM_CHART_NAME}:@$$REPO_DIGEST_OF_ARTIFACT
+	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_CHARTS}/${HELM_CHART_NAME}@$$REPO_DIGEST_OF_ARTIFACT
 
 helm-package:
 	mkdir -p target/helm && helm package --destination target/helm deploy/helm/${OPERATOR_NAME}


template/Makefile.j2
```diff
@@ -59,7 +59,18 @@ docker-publish:
 	fi;\
 	# This generates a signature and publishes it to the registry, next to the image\
 	# Uses the keyless signing flow with Github Actions as identity provider\
-	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:@$$REPO_DIGEST_OF_IMAGE
+	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE;\
+	# Generate the SBOM for the operator image, this leverages the already generated SBOM for the operator binary by cargo-cyclonedx\
+	syft scan -o cyclonedx-json --exclude "/usr/local/bin/stackable-${OPERATOR_NAME}" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}" ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE > sbom.json;\
```
lfrancke, line 64: Sorry for the naive question but....where does the semicolon belong to at the end of the lines?

lfrancke, line 64: - syft also has a -f option to write to a file directly
- I believe syft can "pretty print" output, not sure if that's the default, if it's not I think we should enable it
- With syft you can enable/disable catalogers, maybe we need to disable the cataloger responsible for finding cargo-auditable things to avoid the duplicates

lfrancke (commented):
I haven't checked but I assume shellcheck would complain about missing double quotes everywhere.

Could you run it through shellcheck?

template/Makefile.j2
dervoeti: These commands are basically "one line of bash" but split into several lines (concatenated with `\`, which escapes the newline), hence they need a semicolon between them (could also use `&&` I think). The intention was to be able to write `PURL=...` instead of `export PURL=...`. I should probably do it with `export` though, is prettier now that there are several lines involved. I'll change it.

lfrancke: Can/should we maybe even factor all of this out into a script that you just call here?

dervoeti: There are `make` variables involved, which we would probably need to pass as arguments to the script. So I think it would be possible, but maybe we should create a separate issue for this. It's quite cumbersome to write these Makefiles anyway, since there are a lot of abstractions: I edit the Jinja2 template, a Makefile gets rendered which includes shell script which is executed by Github Actions (which brings in some variables/secrets as environment variables). This sometimes results in difficult escaping of symbols and is hard to test locally, maybe we could find a better solution for our build process (better reproducibility in general would also be nice).

dervoeti: - I added the flag `--output cyclonedx-json=sbom.json` (`--file` is deprecated)
 - By default, Syft spits out compact JSON, which is fine I think, since the SBOM is saved base64 encoded in the OCI registry, so customers need to do some decoding anyway, they don't just download a JSON file. I think how customers see it depends on which tool they use to view the decoded JSON (could also be some GUI / SBOM visualizer). So I'd say: Save it as compact JSON (for efficiency reasons), pretty printing is responsibility of the client. Not a strong opinion though.
 - Sounds good, I'll take a look at the catalogers later

dervoeti: Update on the `export` approach: I could not `make` it work :disappointed: Had problems with variables being out of scope / undefined. I'll leave it for now.

template/Makefile.j2
```diff
@@ -59,7 +59,18 @@ docker-publish:
...
 	# This generates a signature and publishes it to the registry, next to the image\
 	# Uses the keyless signing flow with Github Actions as identity provider\
-	cosign sign -y ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:@$$REPO_DIGEST_OF_IMAGE
+	cosign sign -y "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
+	# Generate the SBOM for the operator image, this leverages the already generated SBOM for the operator binary by cargo-cyclonedx\
+	syft scan --output cyclonedx-json=sbom.json --exclude "/usr/local/bin/stackable-${OPERATOR_NAME}" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}" "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
+	# Determine the PURL for the container image\
+	PURL="pkg:docker/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE?repository_url=${OCI_REGISTRY_HOSTNAME}";\
+	# Get metadata from the image\
+	IMAGE_DESCRIPTION=$$(docker inspect --format='{{.Config.Labels.description}}' ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION});\
```
lfrancke, line 68:
```suggestion
	IMAGE_DESCRIPTION=$$(docker inspect --format='{{.Config.Labels.description}}' "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION}");\
```

template/Makefile.j2
```diff
@@ -59,7 +59,18 @@ docker-publish:
...
+	IMAGE_NAME=$$(docker inspect --format='{{.Config.Labels.name}}' ${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION});\
```
lfrancke, line 69:
```suggestion
	IMAGE_NAME=$$(docker inspect --format='{{.Config.Labels.name}}' "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION}");\
```

lfrancke (changes requested):
I just copied all of it into a shell file and ran shellcheck, no need for it to actually run the script :)

template/.github/workflows/build.yml.j2
```diff
@@ -335,7 +335,10 @@ jobs:
         # Recreate charts and publish charts and docker image. The "-e" is needed as we want to override the
         # default value in the makefile if called from this action, but not otherwise (i.e. when called locally).
         # This is needed for the HELM_REPO variable.
-      - uses: sigstore/cosign-installer@9614fae9e5c5eddabb09f90a270fcb487c9f7149 # tag=v3.3.0
+      - name: Install cosign
+        uses: sigstore/cosign-installer@9614fae9e5c5eddabb09f90a270fcb487c9f7149 # tag=v3.3.0
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0.15.5
```
lfrancke, line 341: Sorry. Just saw this now. Can you also use hash here?

lfrancke (changes requested):

template/Makefile.j2
dervoeti: Disabled the "cargo audit binary" cataloger now, tested it in Linux and MacOS, works as expected. Testrun with airflow-operator was also successful (https://github.com/stackabletech/airflow-operator/actions/runs/7665989842).

lfrancke:
Only tangentially related to this but when I look in Harbor I see a few things not signed by cosign and I have a hard time figuring out what is what.
Is there any way to "see" the SBOMs there?
image

dervoeti:
Yes, I was confused by this in the beginning as well. The thing that's stored in Harbor is not directly the SBOM itself, but an in-toto attestation wrapped in a DSSE envelope (that's the one carrying the signature metadata). These attestations are tagged with the digest of the artifacts it's attached to and a ".att" suffix ("sha256-20f210167078941709d63ec7b7d9b004a95932148ece9a05a9fe73c55d874f0a.att"). This is similar to signatures (there it's a .sig suffix). So the signature is an "attachment" (via this special tag) to the artifact, containing nothing but the data about the signing process and the signature itself. The SBOM attestation artifact contains the signature *plus* the base64 encoded SBOM. Harbor supports displaying signature artifacts that are attached to an OCI artifact (those OCI artifacts are then displayed as "signed"). But the SBOM attestation contains the signature *in itself* (there's no .sig artifact attached to it). That's why it's not displayed as "signed", even though it is (otherwise the cosign verification command would fail). I can try to explain more if needed :slightly_smiling_face: I traversed the whole "attestation verifiaction" flow with `docker manifest inspect` and `oras blob fetch` and decoded the single layers manually.
In the beginning I expected it's simply the SBOM stored as OCI artifact with a separate signature attached to it, which is not the case.

lfrancke:
Pewww....that comment needs to be captured somewhere.

Okay, we can do this as a separate documentation issue, but: We somehow need (I think) an "easy" way for people to download the SBOMs. This can be a separate step where we upload them somewhere for example,

lfrancke (approved):

dervoeti:
Yep. Maybe we can publish them as JSON in separate places (tbd where exactly). I think the current place (Harbor) is more suitable for a "machine-readable" use case: Customers wanting to enforce that all Stackable containers in their cluster have a valid SBOM (signed by one of Stackable's GH Action Workflows) attached. It is possible to get the verified SBOM out of Harbor (with the `cosign` command in the first comment), but this is rather complicated compared to "Just download this JSON file". But it solves the issue of "this JSON file can actually be trusted". If we publish links to JSON files in our release notes, these JSON files are neither signed nor can they in any way directly be related to the build process. The SBOM attestation in Harbor contains signed provenance information along with the SBOM itself:
```
Certificate subject: https://github.com/stackabletech/airflow-operator/.github/workflows/build.yml@refs/heads/spike/attest-sbom
Certificate issuer URL: https://token.actions.githubusercontent.com
GitHub Workflow Trigger: push
GitHub Workflow SHA: 23fadc3b1fa5d9762794af82825b48f4594d8f02
GitHub Workflow Name: Stackable Build Pipeline
GitHub Workflow Repository: stackabletech/airflow-operator
GitHub Workflow Ref: refs/heads/spike/attest-sbom
```

PR #319, opened 2024-02-06
chore: Make it easier for individual operators to upgrade rust before templating

NickLarsenNZ:
This copies how we do it for stackable-cockpit. The previous attempt was reverted in #318 

- `uses` cannot contain expressions, or at least expressions using the `env` context.
- The action hash shouldn't need changing as often as the Rust version.

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -270,6 +270,8 @@ jobs:
           version: v3.13.3
       - name: Set up cargo
         uses: dtolnay/rust-toolchain@be73d7920c329f220ce78e0234b8f96b7ae60248
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN_VERSION }}
       - uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8 # v2.7.1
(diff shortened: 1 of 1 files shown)


template/.github/workflows/build.yml.j2
```diff
@@ -258,7 +269,7 @@ jobs:
         with:
           version: v3.13.3
       - name: Set up cargo
-        uses: dtolnay/rust-toolchain@{[ rust_version }]
+        uses: dtolnay/rust-toolchain@be73d7920c329f220ce78e0234b8f96b7ae60248
```
Techassi, line 272: note: We might want to include an explicit version here just in case.

Techassi (changes requested):
LGTM, just one small note.

template/.github/workflows/build.yml.j2
NickLarsenNZ: Ahh I missed that in my "multi-select" :open_mouth:

NickLarsenNZ: Fixed in 25f1be60eca4f116fc1766a0e4932a34547904b1

Techassi (approved):

PR #323, opened 2024-02-16
chore: Add shell.nix for all operators

NickLarsenNZ:
This is based on what I was doing in stackabletech/secret-operator#344, but to extend it out to all operators.

> [!IMPORTANT]
> Each operator will need to have `Cargo.nix` checked in to be able to make use of these changes. Otherwise they will get the following error when running `nix-shell`:
> ```
> error: getting status of '.../Cargo.nix': No such file or directory
> ```
>
> Run:
> ```sh
> make regenerate-nix
> git add Cargo.nix crate-hashes.json
> git commit -m "add crate2nix generated files"
>
> niv add stackabletech/beku.py
> git add nix/source.json
> git commit -m "add niv dependencies"
> ```

It includes some packages that are only needed for secret-operator:
- They are supposed to already be discovered/provided by default.nix, but I can't get it working.
- If we can't fix the above point, then perhaps we can call out to another nix file that lists additional packages needed by that operator (but then most of then would need to provide a basically empty extra nix file).

---

This will also skip spell-checking in listed generated files.

image

### DIFF
--- a/template/shell.nix.j2
+++ b/template/shell.nix.j2
@@ -20,18 +20,10 @@ in pkgs.mkShell rec {
   ];
 
   # derivation runtime dependencies
-  buildInputs = pkgs.lib.concatMap (crate: crate.buildInputs) cargoDependencySet ++ [ pkgs.libz ];
+  buildInputs = pkgs.lib.concatMap (crate: crate.buildInputs) cargoDependencySet;
   
   # build time dependencies
   nativeBuildInputs = pkgs.lib.concatMap (crate: crate.nativeBuildInputs) cargoDependencySet ++ (with pkgs; [
-    # specific to secret-operator but should be discovered via crate2nix, but
-    # otherwise, we could include an operator-specific nix import for any extras
-    pkg-config
-    clang 
-    openssl
-    protobuf
-    krb5 
-
     git
     yq-go
     jq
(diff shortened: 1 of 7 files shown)


template/shell.nix.j2
```diff
@@ -0,0 +1,44 @@
+let
+  self = import ./. {};
+  inherit (self) sources pkgs;
+
+  beku = pkgs.callPackage (sources."beku.py" + "/beku.nix") {};
+  cargoDependencySetOfCrate = crate: [ crate ] ++ pkgs.lib.concatMap cargoDependencySetOfCrate (crate.dependencies ++ crate.buildDependencies);
+  cargoDependencySet = pkgs.lib.unique (pkgs.lib.flatten (pkgs.lib.mapAttrsToList (crateName: crate: cargoDependencySetOfCrate crate.build) self.cargo.workspaceMembers));
+in pkgs.mkShell rec {
+  name = "{[ operator.name }]";
```
nightkr, line 9: This doesn't need to use Jinja, it can be read from nix/meta.json just like how default.nix does it.

nightkr (approved):
One minor thing, otherwise LGTM.

template/shell.nix.j2
NickLarsenNZ: Ohh yes, good point. now we get syntax highlighting back :D

NickLarsenNZ: Done in 47fc626

nightkr (approved):

NickLarsenNZ:
Moving this back to _In Progress_ to track the needed changes in each generated PR.

PR #327, opened 2024-02-21
ci: add nix install action

NickLarsenNZ:
The operator builds fail when they run `make -e publish` which calls the `build` make target, which then calls `regenerate-nix` target.

@nightkr do we want the build to be `regenerating-nix`? Is it possible that the builds don't then use the versions/hashes in the committed lockfiles?

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -328,6 +328,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           submodules: recursive
+      - uses: cachix/install-nix-action@6004951b182f8860210c8d6f0d808ec5b1a33d28 # tag=v25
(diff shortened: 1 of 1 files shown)


sbernauer (approved):
Approving this as otherwise the build is broken. We can always improve things later on :)

NickLarsenNZ:
@lfrancke I'll leave this in Waiting for Review to give @nightkr  a chance to see my comment in the description (which could lead to a future PR).

If you prefer me to raise an issue so we can move this off the board, let me know :+1:

nightkr:
> @nightkr do we want the build to be regenerating-nix? Is it possible that the builds don't then use the versions/hashes in the committed lockfiles?

crate2nix should respect the Cargo.lock. If anything I'd want CI to fail if regenerate-nix isn't a no-op.

NickLarsenNZ:
> crate2nix should respect the Cargo.lock. If anything I'd want CI to fail if regenerate-nix isn't a no-op.

Ok cool. So the crate2nix and niv lockfile updates are still manually managed.

PR #334, opened 2024-02-28
ci(template): Fork awareness

NickLarsenNZ:
- Remove the `build` target from the `publish` target in Makefile
- Do not run `make -e publish` if is is a PR from a fork (which has no access to secrets)
  - Also don't run the openshift preflight checks which expects there to be an image to pull

---

Tested here:
- https://github.com/stackabletech/hbase-operator/pull/458
- https://github.com/stackabletech/hbase-operator/actions/runs/8078056432

image


fhennig (approved):
looks good!

PR #346, opened 2024-03-17
fix(test): new scripts/run-tests

razvan:
This adds a new helper for running tests called `script/run-tests`.

This script takes the same arguments as the old `scripts/run_tests.sh` but improves
it's behavior in a few ways.

It enables users to run custom CI tests in Jenkins without having to patch
the `release.yaml` file in pull requests.

This is possible given a new argument `--operator <name>=<version>`. When provided,
the operator version specified om the command line is installed instead of
the version present in the `release.yaml` file.

See `./script/run-tests --help` for more information.

It also solves the problem that `stackablectl` doesn't exit with a non-zero code
on failure and therefor the CI pipeline may continue to run with a broken
operator installation.


sbernauer (approved):
Did not review the code properly (as @soenkeliebau assigned himself), but the functionality looks good to me, thanks!

PR #355, opened 2024-04-02
feat: add checks for 'nix' and Kubernetes cluster to 'run-dev' recipe in Makefile

### LINKED ISSUE #335: Give more helpful error messages in 'make run-dev'
[make run-dev](https://github.com/stackabletech/operator-templating/blob/main/template/Makefile.j2#L153C1-L153C8)

- Add a check if nix is installed and helpful message
- Add a check if k8s is reachable - the current error is confusing

fhennig:
closes #335

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -160,5 +160,5 @@ run-dev: check-nix check-kubernetes
 	kubectl apply -f deploy/stackable-operators-ns.yaml
 	nix run -f. tilt -- up --port {[5430 + operator_index}] --namespace stackable-operators
 
-stop-dev: check-nix
+stop-dev: check-nix check-kubernetes
 	nix run -f. tilt -- down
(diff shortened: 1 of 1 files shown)


template/Makefile.j2
```diff
@@ -150,9 +150,15 @@ build: regenerate-charts regenerate-nix helm-package docker-build
...
+
+check-kubernetes:
+	@kubectl cluster-info > /dev/null 2>&1 || (echo "Error: Kubernetes is not running or kubectl is not properly configured."; exit 1)
+
+run-dev: check-nix check-kubernetes
 	kubectl apply -f deploy/stackable-operators-ns.yaml
 	nix run -f. tilt -- up --port {[5430 + operator_index}] --namespace stackable-operators
 
-stop-dev:
+stop-dev: check-nix
```
sbernauer, line 163: As this is also accessing k8s I would argue it would make sense to check it here as well

sbernauer (approved):

template/Makefile.j2
fhennig: done! @sbernauer approve again?

sbernauer (approved):

PR #358, opened 2024-04-29
chore: add arch for preflight

adwk67:
Adds a variable that selects the target architecture, as is done for docker-images https://github.com/stackabletech/docker-images/pull/626. N.B. preflight does not expect a `linux/` prefix but just `amd64` or `arm64`.
Fixes https://github.com/stackabletech/issues/issues/559.
Tested here:https://github.com/stackabletech/hive-operator/pull/450/files.
Dry-run for PRs: https://github.com/stackabletech/operator-templating/actions/runs/8881229144

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -376,6 +376,8 @@ jobs:
           wget https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/download/1.9.1/preflight-linux-amd64
           chmod +x preflight-linux-amd64
       - name: Check container
-        run: ./preflight-linux-amd64 check container "$IMAGE_TAG" > preflight.out
+        run: |
+          ARCH_FOR_PREFLIGHT="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
+          ./preflight-linux-amd64 check container "$IMAGE_TAG" --platform "${ARCH_FOR_PREFLIGHT}" > preflight.out
       - name: "Passed?"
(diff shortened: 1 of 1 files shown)


template/.github/workflows/build.yml.j2
```diff
@@ -376,6 +376,8 @@ jobs:
           wget https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/download/1.9.1/preflight-linux-amd64
           chmod +x preflight-linux-amd64
       - name: Check container
-        run: ./preflight-linux-amd64 check container "$IMAGE_TAG" > preflight.out
+        run: |
+          ARCH_FOR_PREFLIGHT="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
+          ./preflight-linux-amd64 check container "$IMAGE_TAG" --platform "${ARCH_FOR_PREFLIGHT}" > preflight.out
```
sbernauer, line 381: The `amd64` suffix in  `/preflight-linux-amd64` seems weird, but when it works I'm happy ^^

sbernauer (approved):

sbernauer:
Approving assuming the test with hive was successful

template/.github/workflows/build.yml.j2
adwk67: Thanks for approving! Yes, preflight seems to supply the prefix itself. If we pass `linux/amd64` it fails with:
`preflight .../stackable-experimental/hive:3.1.3-stackable24.3.0 --platform linux/amd64 --> no child with platform linux/linux/amd64`

PR #378, opened 2024-05-10
Multi-Architecture Operator pipeline

Maleware:
This adds:
- build.yml.j2 updated workflow to use new cargo-build-action as well as leveraging BuildJet runner to natively support arm64
- Makefile.j2 Adding creation of manifest llists as well as sufficient tags for multi architecture releases of stackable operators

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -426,4 +426,4 @@ jobs:
           ARCH_FOR_PREFLIGHT="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
           ./preflight-linux-amd64 check container "$IMAGE_TAG" --platform "${ARCH_FOR_PREFLIGHT}" > preflight.out
       - name: "Passed?"
-        run: '[ "$(./preflight-linux-amd64 check container "$IMAGE_TAG" --platform "${ARCH_FOR_PREFLIGHT}" | jq -r .passed)" == true ]'
+        run: '[ "$(cat preflight.out | jq -r .passed)" == true ]'
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -391,9 +391,9 @@ jobs:
         uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
         with:
           submodules: recursive
-        # This step checks if the current run was triggered by a push to a pr (or a pr being created).
-        # If this is the case it changes the version of this project in all Cargo.toml files to include the suffix
-        # "-pr<prnumber>" so that the published artifacts can be linked to this PR.
+      # This step checks if the current run was triggered by a push to a pr (or a pr being created).
+      # If this is the case it changes the version of this project in all Cargo.toml files to include the suffix
+      # "-pr<prnumber>" so that the published artifacts can be linked to this PR.
       - uses: stackabletech/cargo-install-action@main
         with:
           crate: cargo-edit
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -220,7 +220,7 @@ jobs:
       - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
         with:
           submodules: recursive
-      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # tag=v5.1.0
+      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: '3.12'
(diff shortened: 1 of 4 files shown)


Maleware:
Solving: https://github.com/stackabletech/issues/issues/553

template/.github/workflows/build.yml.j2
```diff
@@ -340,15 +343,24 @@ jobs:
...
+        run: |
+          # Installing helm on BuildJet only
+          if [ "$(arch)" = "aarch64" ]; then
+            curl https://baltocdn.com/helm/signing.asc | gpg --dearmor | sudo tee /usr/share/keyrings/helm.gpg > /dev/null
+            sudo apt-get -y install apt-transport-https --yes
+            echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/helm.gpg] https://baltocdn.com/helm/stable/debian/ all main" | sudo tee /etc/apt/sources.list.d/helm-stable-debian.list
+            sudo apt-get -y update
+            sudo apt-get -y install helm
+          fi
+          make -e build
```
sbernauer, line 363: Nit
```suggestion

          make -e build
```

template/.github/workflows/build.yml.j2
```diff
@@ -376,4 +425,4 @@ jobs:
           ARCH_FOR_PREFLIGHT="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
           ./preflight-linux-amd64 check container "$IMAGE_TAG" --platform "${ARCH_FOR_PREFLIGHT}" > preflight.out
```
sbernauer, line 426: Still confused about the `amd64` suffix, but if it works...

template/.gitignore.j2
```diff
@@ -17,3 +17,5 @@ tilt_options.json
 .direnv/
 .direnvrc
 .envrc
+
+.DS_Store
```
sbernauer, line 21: LGTM, but just FYI https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

template/Makefile.j2
```diff
@@ -76,13 +77,39 @@ docker-publish:
...
+	IMAGE_DESCRIPTION=$$(docker inspect --format='{{.Config.Labels.description}}' "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION}-${ARCH}");\
+	IMAGE_NAME=$$(docker inspect --format='{{.Config.Labels.name}}' "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION}-${ARCH}");\
 	# Merge the SBOM with the metadata for the operator\
 	jq -s '{"metadata":{"component":{"description":"'"$$IMAGE_NAME. $$IMAGE_DESCRIPTION"'","supplier":{"name":"Stackable GmbH","url":["https://stackable.tech/"]},"author":"Stackable GmbH","purl":"'"$$PURL"'","publisher":"Stackable GmbH"}}} * .[0]' sbom.json > sbom.merged.json;\
 	# Attest the SBOM to the image\
 	cosign attest -y --predicate sbom.merged.json --type cyclonedx "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE"
 
+# This assumes "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:${VERSION}-amd64 and "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:${VERSION}-arm64 being build and pushed
```
sbernauer, line 87:
```suggestion
# This assumes "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:${VERSION}-amd64 and "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}:${VERSION}-arm64 are build and pushed
```

sbernauer (changes requested):
Only nits

template/.github/workflows/build.yml.j2
Maleware: This is not selecting the image, it's a hint for `./preflight-linux-amd64` which architecture it's checking on. However, since this is running only on amd64 it will only check this at this moment. @adwk67 would you rather like to see this running on `arm64` and `amd64`?

sbernauer (approved):
LGTM % remaining question

template/.github/workflows/build.yml.j2
adwk67: I don't think it is that important to have preflight running on different platforms, but rather just to make sure that preflight works consistently, and the platform flag just makes sure that preflight pulls an image that matches the architecture where it is running (although it is also seems to default to the current platform _anyway_).

PR #380, opened 2024-05-10
fix: remove nix from CI

razvan:


NickLarsenNZ (commented):
What problem was it causing?
Just extra time, or failing?

I guess we should update niv/crate2nix when we do other version bumps. Should we add that into the developer docs (I am happy to do it)?

NickLarsenNZ (approved):
Will approve anyway as I don't think there is a negative consequence here.

NickLarsenNZ:
Please don't roll out until the beku.py PR is merged as I want to include the fix (unless you urgently need this). Saves a bunch of time merging PRs.

razvan:
ok, thanks

nightkr:
Also curious here, generally `regenerate-charts` is the one that requires the operator rebuild. This will cause the Nix lockfiles to drift out of sync with the Cargo ones.

PR #381, opened 2024-05-13
chore(nix): Update dependencies

NickLarsenNZ:
- Use fixed beku version from stackabletech/beku.py#23
- Update crate2nix and nixpkgs


Maleware (approved):
LGTM

PR #385, opened 2024-05-15
chore: Use uname instead of arch in Makefile

sbernauer:
[NixOS does not ship these by default](https://github.com/coreutils/coreutils/blob/684db6cb9b06cbb4f246152da427da6f47aa16cc/build-aux/gen-lists-of-programs.sh#L12-L22) and arch is only a wrapper around `uname -m` anyway


nightkr:
To clarify, this isn't just a NixOS issue, arch is disabled by default _by the upstream_. It's not in Arch either, for example.

nightkr (approved):

PR #387, opened 2024-05-29
Switching to ubicloud arm runner

Maleware:
Adds ubicloud-standard-8-arm to operator build pipeline

### DIFF
--- a/template/.actionlint.yaml
+++ b/template/.actionlint.yaml
@@ -2,5 +2,4 @@
 self-hosted-runner:
   # BuildJet machines we are using
   labels:
-    - buildjet-2vcpu-ubuntu-2204-arm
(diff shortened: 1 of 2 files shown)


template/.actionlint.yaml
```diff
@@ -2,5 +2,4 @@
 self-hosted-runner:
   # BuildJet machines we are using
```
razvan, line 3: I guess this comment is now wrong

razvan (approved):
see comment

template/.actionlint.yaml
Maleware: Thank you! Good catch :)

Maleware:
https://github.com/stackabletech/opa-operator/actions/runs/9596294742

tested successfully.

razvan (approved):
lgtm!

PR #390, opened 2024-06-13
feat: add workflow_dispatch and schedule event triggers to the operator build workflow

xeniape:
This PR adds the workflow_dispatch and schedule event triggers to run the operator image build manually if needed and to have a scheduled build once a week on the weekend.

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -68,11 +68,11 @@ jobs:
   #        repository: test
   #
   #    - all tagged releases land in stable:
-  #         condition: github.event_name == 'create' & github.ref.startswith('refs/tags/')
+  #         condition: github.event_name == 'push' & github.ref.startswith('refs/tags/')
   #         repository: stable
   #
-  #    - all pushes to main (i.e. PR-merges) land in dev:
-  #         condition: github.event_name == 'push' & github.ref == 'refs/heads/main'
+  #    - all pushes to main (i.e. PR-merges) and all scheduled/manual workflow runs on main land in dev:
+  #         condition: ( github.event_name == 'push' | github.event_name == 'schedule' | github.event_name == 'workflow_dispatch' ) & github.ref == 'refs/heads/main'
   #         repository: dev
   #
   # Any other scenarios (e.g. when a branch is created/pushed) will cause the publish step to be skipped, most commonly this is expected to happen for the
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -16,6 +16,9 @@ on:
       - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
   merge_group:
+  schedule:
+    - cron: '15 3 * * 6'
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -88,13 +88,13 @@ jobs:
           TRIGGER: ${{ github.event_name }}
           GITHUB_REF: ${{ github.ref }}
         run: |
-          if [[ $TRIGGER == "pull_request" || ( $TRIGGER == "workflow_dispatch" && $GITHUB_REF != refs/heads/main )]]; then
+          if [[ $TRIGGER == "pull_request" ]]; then
             echo "exporting test as target helm repo: ${{ env.TEST_REPO_HELM_URL }}"
             echo "helm_repo=${{ env.TEST_REPO_HELM_URL }}" >> $GITHUB_OUTPUT
           elif [[ ( $TRIGGER == "push" || $TRIGGER == "schedule" || $TRIGGER == "workflow_dispatch" ) && $GITHUB_REF == "refs/heads/main" ]]; then
             echo "exporting dev as target helm repo: ${{ env.DEV_REPO_HELM_URL }}"
             echo "helm_repo=${{ env.DEV_REPO_HELM_URL }}" >> $GITHUB_OUTPUT
-          elif [[ ( $TRIGGER == "push" ) && $GITHUB_REF == refs/tags/* ]]; then
+          elif [[ $TRIGGER == "push" && $GITHUB_REF == refs/tags/* ]]; then
             echo "exporting stable as target helm repo: ${{ env.STABLE_REPO_HELM_URL }}"
             echo "helm_repo=${{ env.STABLE_REPO_HELM_URL }}" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (dismissed):
LGTM

template/.github/workflows/build.yml.j2
```diff
@@ -16,6 +16,9 @@ on:
       - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
   merge_group:
+  schedule:
+    - cron: '15 3 * * 6'
+  workflow_dispatch:
```
sbernauer, line 21: Below (line 90) is a script that checks why the run was triggered. Is this already adopted to work with the new trigger types?

template/.github/workflows/build.yml.j2
```diff
@@ -16,6 +16,9 @@ on:
       - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
   merge_group:
+  schedule:
```
lfrancke, line 19: If possible I'd like to have a comment here for those like me that are crontabwise challenged. I can never remember the syntax.

Can we have a comment saying that this runs every saturday night?

lfrancke (changes requested):

template/.github/workflows/build.yml.j2
NickLarsenNZ: Or a link to the crontabguru explanation (or is that too much indirection?)

https://crontab.guru/#15_3_*_*_6

template/.github/workflows/build.yml.j2
NickLarsenNZ: Good spotting:

https://github.com/stackabletech/operator-templating/blob/259470bab1b67b372048fdae5aaf2fb34754e8bc/template/.github/workflows/build.yml.j2#L96-L99

template/.github/workflows/build.yml.j2
lfrancke: Either is fine for me I think

template/.github/workflows/build.yml.j2
NickLarsenNZ: Thinking about this a little more, maybe can we ignore publishing charts completely for these new triggers:
* on `schedule`: The purpose was to only rebuild the image to catch any security updates
* on `workflow_dispatch`: Manually triggering is more ambiguous, becuase the _users_ intention might not be for rebuilding the image. Would we want to publish helm charts when the GITHUB_REF is on the following?
    * for `refs/heads/main`, the intention is probably to rebuild an image earlier than the next schedule, so no need to publish charts.
    * for `refs/heads/*`, Unknown intent, probably better taken care of by a `pull_request` and the existing conditions.
    * for `refs/tags/*`, probably should not be run? IMO, these should be static for a release, so push a new tag instead (unless we nede to consider non-release tags too.

Regarding image builds, would we want to avoid build/publish on `workflow_dispatch` for `refs/tags/*`?

template/.github/workflows/build.yml.j2
```diff
@@ -342,6 +346,9 @@ jobs:
       - name: Update version if PR
         if: ${{ github.event_name == 'pull_request' }}
         run: cargo set-version --offline --workspace 0.0.0-pr${{ github.event.pull_request.number }}
+      - name: Update version if manual trigger on non-main branch
```
sbernauer, line 349: I think I would prefer if you could only run a workflow dispatch on `main`. Everything else introduces new image variants that currently don't exist (think of cleanup jobs and similar). Also we all have permissions to push an empty commit to all branches beside main (which also triggers ci)

Maleware: I agree on that. Afaik this pipeline runs on PR, thus this use case would be already covered I think.

template/.github/workflows/build.yml.j2
```diff
@@ -87,10 +91,10 @@ jobs:
           if [[ $TRIGGER == "pull_request" ]]; then
             echo "exporting test as target helm repo: ${{ env.TEST_REPO_HELM_URL }}"
             echo "helm_repo=${{ env.TEST_REPO_HELM_URL }}" >> $GITHUB_OUTPUT
-          elif [[ $TRIGGER == "push" && $GITHUB_REF == "refs/heads/main" ]]; then
+          elif [[ ( $TRIGGER == "push" || $TRIGGER == "schedule" || $TRIGGER == "workflow_dispatch" ) && $GITHUB_REF == "refs/heads/main" ]]; then
             echo "exporting dev as target helm repo: ${{ env.DEV_REPO_HELM_URL }}"
             echo "helm_repo=${{ env.DEV_REPO_HELM_URL }}" >> $GITHUB_OUTPUT
-          elif [[ ( $TRIGGER == "create" || $TRIGGER == "push" ) && $GITHUB_REF == refs/tags/* ]]; then
```
xeniape, line 93: I removed the check for the "create" event as part of this PR as well.
This event does not appear in the `on:` configuration and therefore does not get triggered (anymore?) by created branches or tags. Tested the setup in a dummy-repo and the dummy workflow was not triggered by a new branch/tag, so it seems to be safe to remove for me.

But happy to hear other opinions on removing it.

lfrancke (dismissed):
I approve but I have only looked at my comment :)

sbernauer (approved):
Change looks good and minimal to me :+1:

template/.github/workflows/build.yml.j2
NickLarsenNZ: Marking this as resolved, as I believe future changes cover these items.

template/.github/workflows/build.yml.j2
```diff
@@ -84,13 +88,13 @@ jobs:
           TRIGGER: ${{ github.event_name }}
           GITHUB_REF: ${{ github.ref }}
         run: |
-          if [[ $TRIGGER == "pull_request" ]]; then
+          if [[ $TRIGGER == "pull_request" || ( $TRIGGER == "workflow_dispatch" && $GITHUB_REF != refs/heads/main )]]; then
```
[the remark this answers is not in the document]
NickLarsenNZ: Obsolete comment

template/.github/workflows/build.yml.j2
```diff
@@ -342,6 +346,9 @@ jobs:
       - name: Update version if PR
         if: ${{ github.event_name == 'pull_request' }}
         run: cargo set-version --offline --workspace 0.0.0-pr${{ github.event.pull_request.number }}
+      - name: Update version if manual trigger on non-main branch
+        if: ${{ github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main' }}
+        run: cargo set-version --offline --workspace branch-${{ github.ref_name }}
```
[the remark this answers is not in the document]
NickLarsenNZ: Obsolete comment

template/.github/workflows/build.yml.j2
```diff
@@ -16,6 +16,10 @@ on:
       - '[0-9][0-9].[0-9]+.[0-9]+'
   pull_request:
   merge_group:
+  # rebuild of main branch runs every saturday night
+  schedule:
+    - cron: '15 3 * * 6'
+  workflow_dispatch:
```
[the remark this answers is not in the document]
NickLarsenNZ: Done in a71dea5

template/.github/workflows/build.yml.j2
NickLarsenNZ, line 1: This comment also needs updating
https://github.com/stackabletech/operator-templating/blob/a71dea562c3ff5ae0622e6f2decd8018a7402881/template/.github/workflows/build.yml.j2#L62-L77

xeniape: Updated in [63ed92b](https://github.com/stackabletech/operator-templating/pull/390/commits/63ed92b3b2d68ba2b951c69cdef856fef74dc722)

NickLarsenNZ (approved):
LGTM

PR #420, opened 2024-08-16
chore: Bump to Rust 0.80.1 (finally)

sbernauer:
This will fix `make run-dev` for operators already using Rust 1.80 language features (such as druid) and unblock us from using them everywhere (e.g. in operator-rs)


NickLarsenNZ (approved):
LGTM

PR #430, opened 2024-09-13
Bump versions

lfrancke:
This updates (amongst other things) cargo-deny to 0.16.1 which includes breaking changes.
This PR includes the updated `deny.toml`

Previously we would [warn](https://github.com/stackabletech/operator-templating/pull/430/files#diff-35350462e19fc0bd4fab36c76e0ee1ba731c7580f9f3dc1a25fc9fb170cd1a73) on vulnerabilities, unmaintained crates etc.. This functionality has been removed in cargo-deny:
```
The following fields have all been removed in favor of denying all advisories by default. To ignore an advisory the ignore field can be used as before.

vulnerability - Vulnerability advisories are now deny by default
unmaintained - Unmaintained advisories are now deny by default
unsound - Unsound advisories are now deny by default
notice - Notice advisories are now deny by default
```

I did run the new config on hdfs operator and there were no failures but I fully expect this to cause issues in some other operator. I'll be there to clean it up and add all relevant advisories to the ignore list or fix the operators.

What I'd like to do is to merge this in operator-templating and kick off a run and then see all the broken PRs and fix them in a follow-up PR in operator-templating. If that is not okay for anyone please let me know and I'll check all operators manually before merging.
This is especially the case if anyone needs anything merged via operator-templating today or tomorrow...

### DIFF
--- a/template/deny.toml
+++ b/template/deny.toml
@@ -1,3 +1,4 @@
+[graph]
 targets = [
     { triple = "x86_64-unknown-linux-gnu" },
     { triple = "aarch64-unknown-linux-gnu" },
@@ -7,20 +8,13 @@ targets = [
 ]
 
 [advisories]
-vulnerability = "warn"
-unmaintained = "allow"
-unsound = "warn"
-yanked = "warn"
-notice = "warn"
+yanked = "deny"
 
 [bans]
 multiple-versions = "allow"
 
 [licenses]
-unlicensed = "deny"
-copyleft = "deny"
-allow-osi-fsf-free = "neither"
-default = "deny"
+unused-allowed-license = "allow"
 confidence-threshold = 1.0
 allow = [
     "Apache-2.0",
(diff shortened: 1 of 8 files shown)


template/deny.toml
```diff
@@ -7,20 +8,13 @@ targets = [
 ]
 
 [advisories]
-vulnerability = "warn"
-unmaintained = "allow"
```
Techassi, line 11: note: Just making sure this is deliberate. Removing `unmaintained` here will now default to `deny`. Is this the new intended behaviour?

Techassi (approved):
LGTM.

One quick note: The action in version `2.0.1` uses cargo-deny `0.16.1`. It might be worth to update the PR description to reflect that.

template/deny.toml
lfrancke: Yes. It is. According to [cargo-deny](https://github.com/EmbarkStudios/cargo-deny/releases/tag/0.16.0) at least.

We'll have to manually list all _unmaintained_ advisories in the _ignore_ section now.
I tried it with the hdfs operator and didn't get any warnings but you're correct: I should announce when this happens and I'll be there for the cleanup.

PR #431, opened 2024-09-13
Configure renovate so that it should be able to pick up pre-commit

lfrancke:
`pre-commit` support is at the moment only in beta test for renovate so it has to be enabled manually in the config for renovate. I'll do that separately.

https://docs.renovatebot.com/modules/manager/pre-commit/


Techassi (approved):

lfrancke:
Turns out renovate does not support digest updating for pre-commit yet: https://github.com/renovatebot/renovate/issues/22567

PR #432, opened 2024-09-13
Dockerfile updates

lfrancke:
- Also uses OpenContainer annotations as in our product images
- Switches to a different user id for security reasons

### DIFF
--- a/template/docker/Dockerfile.j2
+++ b/template/docker/Dockerfile.j2
@@ -9,23 +9,78 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal AS operator
 ARG VERSION
 ARG RELEASE="1"
 
-LABEL name="Stackable Operator for {[ operator.pretty_string }]" \
-  maintainer="info@stackable.tech" \
-  vendor="Stackable GmbH" \
-  version="${VERSION}" \
-  release="${RELEASE}" \
-  summary="Deploy and manage {[ operator.pretty_string }] clusters." \
-  description="Deploy and manage {[ operator.pretty_string }] clusters."
+# These are chosen at random and are this high on purpose to have very little chance to clash with an existing user or group on the host system
+ARG STACKABLE_USER_GID="574654813"
+ARG STACKABLE_USER_UID="782252253"
+
+# These labels have mostly been superseded by the OpenContainer spec annotations below but it doesn't hurt to include them
+# http://label-schema.org/rc1/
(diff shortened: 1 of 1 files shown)


lfrancke:
Pre-commit is wrong. The generated Dockerfiles look good to me.

razvan:
> Pre-commit is wrong. The generated Dockerfiles look good to me.

It's the wrong hadolint version. It should be 2.12.0

Techassi:
Yes, I updated it in https://github.com/stackabletech/operator-templating/pull/435

lfrancke:
Okay...this is ready for review again. The remaining hadolint warning is not a new one and occurs due to the Jinja syntax in the Dockerfile template

template/docker/Dockerfile.j2
```diff
@@ -1,31 +1,90 @@
...
-LABEL name="Stackable Operator for {[ operator.pretty_string }]" \
-  maintainer="info@stackable.tech" \
-  vendor="Stackable GmbH" \
-  version="${VERSION}" \
-  release="${RELEASE}" \
-  summary="Deploy and manage {[ operator.pretty_string }] clusters." \
-  description="Deploy and manage {[ operator.pretty_string }] clusters."
+# These are chosen at random and are this high on purpose to have very little chance to clash with an existing user or group on the host system
+ARG STACKABLE_USER_GID="574654813"
+ARG STACKABLE_USER_UID="782252253"
```
sbernauer, line 18: @Techassi can you please give some details on the high ID numbers on some Linux systems?

lfrancke: I did some research on this and it seems as if "in the old days" only 16 bit UIDs were supported but that restriction seems to have been lifted a long time ago.

I'd say we're good with having this in.

https://unix.stackexchange.com/questions/191663/any-concerns-with-using-high-uid-numbers-3000-on-rhel5

sbernauer (dismissed):
LGTM, thanks!

template/docker/Dockerfile.j2
sbernauer: Works for me

lfrancke:
Thank you. I just need to resolve the conflicts.

sbernauer (approved):

lfrancke:
Unfortunately, this has a bug. It references an argument `STACKABLE_USER_NAME` that doesn't exist. I'll test and create a followup PR

PR #439, opened 2024-09-18
Fix broken Dockerfile and minor improvements

lfrancke:
This fixes an issue introduced by a past PR (https://github.com/stackabletech/operator-templating/pull/432). It uses an argument that doesn't exist ("STACKABLE_USER_NAME").

In the course of diagnosing that I stumbled across a few inconsistencies with our "docker-images" Dockerfiles which I am addressing here as well.

- Updates the Helm version used (not caught by Renovate yet)
- Sets a build timeout for the build & package phase to two hours (down from the default of 6)
- Adds the dnf.conf file
- Adds convenience utilities
- Adds a few common aliases and a better prompt

### DIFF
--- a/template/docker/Dockerfile.j2
+++ b/template/docker/Dockerfile.j2
@@ -66,23 +66,55 @@ LABEL io.openshift.tags="ubi9,stackable,sdp,{[ operator.product_string }]"
 LABEL io.k8s.description="Deploy and manage {[ operator.pretty_string }] clusters."
 LABEL io.k8s.display-name="Stackable Operator for {[ operator.pretty_string }]"
 
+COPY <<EOF /etc/dnf/dnf.conf
+[main]
+install_weak_deps=0
+assumeyes=True
+tsflags=nodocs
+EOF
+
 RUN <<EOF
-# Update image and install kerberos client libraries
-# install_weak_deps in microdnf does not support the literal "False" as dnf does
-# https://github.com/rpm-software-management/microdnf/blob/a600c62f29262d71a6259b70dc220df65a2ab9b5/dnf/dnf-main.c#L176-L189
+# Update image and install kerberos client libraries as well as some other utilities
 microdnf update
+
+# **iputils**
+# To make debugging easier, includes things like ping
+# Added 2024-03: We cannot find any vulnerabilities in the past years
+# https://github.com/iputils/iputils
+#
+# **less**
+# To make debugging easier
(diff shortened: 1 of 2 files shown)


template/docker/Dockerfile.j2
```diff
@@ -66,23 +66,55 @@ LABEL io.openshift.tags="ubi9,stackable,sdp,{[ operator.product_string }]"
...
 RUN <<EOF
-# Update image and install kerberos client libraries
-# install_weak_deps in microdnf does not support the literal "False" as dnf does
-# https://github.com/rpm-software-management/microdnf/blob/a600c62f29262d71a6259b70dc220df65a2ab9b5/dnf/dnf-main.c#L176-L189
+# Update image and install kerberos client libraries as well as some other utilities
 microdnf update
+
+# **iputils**
+# To make debugging easier, includes things like ping
+# Added 2024-03: We cannot find any vulnerabilities in the past years
```
sbernauer, line 82: Was this really already added in 2024-03? :)

sbernauer (approved):
LGTM otherwise

template/docker/Dockerfile.j2
lfrancke: It was over in the docker-images repo, yeah

sbernauer: Hmm :thinking: Same with 24.3.0
```
docker run -it --entrypoint bash docker.stackable.tech/stackable/trino-operator:24.7.0                                                                     kind-kind
[stackable@7374d513b091 /]$ ping
bash: ping: command not found
[stackable@7374d513b091 /]$ less
bash: less: command not found
```

sbernauer: But honestly I don't care ;)

lfrancke: ah! Sorry. It was added in March 2024, not necessarily for the 24.3 release

It should be in 24.7 though

sbernauer: It's in neither but anyway ;)

lfrancke: Uhm...that's weird. I'll take a look at a later time. Thanks.

sbernauer: Just for the record `docker run -it --entrypoint ping docker.stackable.tech/stackable/trino-operator:24.3.0`

lfrancke: Ahhh! It has _not_ been added to operators before, only products up until now.

PR #440, opened 2024-09-19
ci(template): Ignore build for doc changes

NickLarsenNZ:
It makes little sense to build everything if only docs have changed.

> [!CAUTION]
> Need to check if the "Stackable Build Pipeline" is a required check in Github Repo Settings. Would a PR with the build job skipped be allowed to merge?


NickLarsenNZ:
Apparently this will break the required checks.

So we would then need an initial job to see if there are any changes outside of docs, and then continue.

Probably needs `if` and `needs` on each of the subsequent jobs to control execution.

PR #443, opened 2024-10-10
chore: Migrate pre-commit config

Techassi:
Note: This requires pre-commit version 4.0.0 or later. Earlier versions will fail on the new stage names.


NickLarsenNZ (approved):
LGTM

PR #446, opened 2024-10-15
feat: run-tests can skip ops during release installation

razvan:
~~'make run-dev' now installs ops in 'release.yaml' together with the operator installed by Tilt.~~

Add `--skip-operator` option to the run tests script. The use case is to skip ops installed  by Tilt.

### DIFF
--- a/template/scripts/run-tests
+++ b/template/scripts/run-tests
@@ -217,7 +217,7 @@ def release_file(
                             patch_version = version
                             ops_copy.remove((op, version))  # found an operator to patch
                             break
-                patched_release.append(line.rstrip('\n'))
+                patched_release.append(line.rstrip("\n"))
 
         # Filter out skip operators
         skipped_release = []
--- a/template/scripts/run-tests
+++ b/template/scripts/run-tests
@@ -95,7 +95,7 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
     parser.add_argument(
         "--operator",
         help="Patch operator version in release.yaml. Format <operator>=<version>",
-        nargs="*",
+        action="append",
         type=cli_parse_operator_args,
         default=[],
     )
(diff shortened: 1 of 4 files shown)


nightkr:
Strongly -1 for me for this one, I don't want `run-dev` to mess with the rest of the cluster.

razvan:
> Strongly -1 for me for this one, I don't want `run-dev` to mess with the rest of the cluster.

I see. I updated the PR.

template/scripts/run-tests
```diff
@@ -209,7 +217,22 @@ def release_file(
...
+            skip = False
+            if skip:
+                skip = False
+            else:
+                if any((op in line for op in skip_ops)):
+                    skip = True
+                    break
+                else:
+                    skipped_release.append(line)
+        patched_release = skipped_release
```
nightkr, line 234: I would have assumed something like `patched_release = "\n".join([line for line in patched_release if not any((op in line for op in skip_ops))])`. That said, this kind of string mangling of the YAML seems really fragile. Why not use a proper YAML parser instead?

razvan: Your suggestion won't work because it needs to skip more than just one line.

> Why not use a proper YAML parser instead?

Seemed overkill back then. Maybe it's more appropriate now.

razvan: > I'm confused, the PR seems to abort as soon as it encounters any skipped op?

You are absolutely correct. You pointed out to a bug that was also affecting the `--operator` argument. I fixed them both now.

Also I would love to use a yaml parser now , but before we do that, I want to move this script to a proper installable package.

Until then, please bare with me.

nightkr: Alright, that's fair.

razvan:
@nightkr is there anything else I can do here ?

nightkr (commented):
Oh I forgot to submit the review >.<

nightkr (approved):

razvan:
thanks!

PR #448, opened 2024-10-16
fix: make PURLs use oci type / include architecture in `--source-version`

dervoeti:
Same thing as https://github.com/stackabletech/actions/pull/11, but for our operator images. It probably makes sense for the same person to review both this PR and the other.

---

Use OCI type for PURLs, according to https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci

This is more correct in general and most importantly makes the PURL equal to the one Trivy generates for our container images.

Hint: To check the PURL Trivy generates, run for example these commands:
```
trivy image --scanners vuln -f cyclonedx -o airflow-result.cdx.json oci.stackable.tech/sdp/airflow-operator:0.0.0-dev-amd64
jq -r '.metadata.component.purl' airflow-result.cdx.json
```
---
The second commit includes the architecture in the component version in the SBOM (`0.0.0-dev-amd64` instead of `0.0.0-dev`). That is in line with the image tag and we also do it this way with our product images.

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -50,7 +50,8 @@ docker-publish:
 	# Generate the SBOM for the operator image, this leverages the already generated SBOM for the operator binary by cargo-cyclonedx\
 	syft scan --output cyclonedx-json=sbom.json --select-catalogers "-cargo-auditable-binary-cataloger" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}" "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
 	# Determine the PURL for the container image\
-	PURL="pkg:docker/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE?repository_url=${DOCKER_REPO}";\
+	URLENCODED_REPO_DIGEST_OF_IMAGE=$$(echo "$$REPO_DIGEST_OF_IMAGE" | sed 's/:/%3A/g');\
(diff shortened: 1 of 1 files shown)


template/Makefile.j2
```diff
@@ -48,9 +48,10 @@ docker-publish:
 	# Uses the keyless signing flow with Github Actions as identity provider\
 	cosign sign -y "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
 	# Generate the SBOM for the operator image, this leverages the already generated SBOM for the operator binary by cargo-cyclonedx\
-	syft scan --output cyclonedx-json=sbom.json --select-catalogers "-cargo-auditable-binary-cataloger" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}" "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
+	syft scan --output cyclonedx-json=sbom.json --select-catalogers "-cargo-auditable-binary-cataloger" --scope all-layers --source-name "${OPERATOR_NAME}" --source-version "${VERSION}-${ARCH}" "${DOCKER_REPO}/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE";\
 	# Determine the PURL for the container image\
-	PURL="pkg:docker/${ORGANIZATION}/${OPERATOR_NAME}@$$REPO_DIGEST_OF_IMAGE?repository_url=${DOCKER_REPO}";\
+	URLENCODED_REPO_DIGEST_OF_IMAGE=$$(echo "$$REPO_DIGEST_OF_IMAGE" | sed 's/:/%3A/g');\
```
Techassi, line 53: Are we sure that this covers all possible inputs and ensures the value is url-encoded correctly? We could instead use

```shell
$(echo "$REPO_DIGEST_OF_IMAGE" | jq -Rr @uri)
```

dervoeti: Yeah it should be fine, I can't see a way how any other special character can be in there. And even if, we'd have to look into why that's the suddenly case.

PR #455, opened 2024-10-29
fix(playbook): Retired files and local testing improvements

NickLarsenNZ:
- Use correct filename, and add missing file to the retired files list.
- Refactor the playbook to run more steps in the local test.
- Pull in a change from stackabletech/opa-operator#622.

### DIFF
--- a/playbook/update_repo.yaml
+++ b/playbook/update_repo.yaml
@@ -111,6 +111,16 @@
       register: git_change
       failed_when: git_change.rc == 0 # fail when there are no changes
 
+    - name: Configure git author mail
+      command:
+        argv: [git, config, user.email, "{{ author_mail }}"]
+        chdir: "{{ work_dir }}/{{ operator.name }}"
+
+    - name: Configure git author name
+      command:
+        argv: [git, config, user.name, "{{ author_name }}"]
+        chdir: "{{ work_dir }}/{{ operator.name }}"
+
     - name: "Operator [{{ operator.name }}] stage all changes"
       command:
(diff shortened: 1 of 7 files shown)


playbook/update_repo.yaml
```diff
@@ -112,26 +112,30 @@
       failed_when: git_change.rc == 0 # fail when there are no changes
 
 
-# Check if anything was changed, if yes commit changes and create a pull request, otherwise skip rest of this play
+# Commit changes
```
soenkeliebau: I was proud of my comment actually stating that this influences further execution of the playbook ':)

template/default.nix
```diff
@@ -16,6 +16,10 @@
       stackable-secret-operator = attrs: {
         buildInputs = [ pkgs.protobuf pkgs.rustfmt ];
       };
+      stackable-opa-user-info-fetcher = attrs: {
+        # TODO: why is this not pulled in via libgssapi-sys?
+        buildInputs = [ pkgs.krb5 ];
+      };
```
Techassi, line 22: note(non-blocking): We should only put common stuff in the templated `default.nix`. All other operator-specific stuff should _not_ be templated and should live in the operator repositories. @NickLarsenNZ will look into that.

Techassi (dismissed):

template/default.nix
NickLarsenNZ: Issue created here: https://github.com/stackabletech/operator-templating/issues/456

Techassi (approved):

PR #470, opened 2024-12-04
ci: Workflow changes in preparation for release-candidates

adwk67:
See https://github.com/stackabletech/stackable-utils/pull/95/files

This PR:

- extends the tag regex to allow optional `-rcXX` suffixes
- suppresses the automatic setting of the version when merging a PR to a branch other than `main`. The version should be set in the PR creation (e.g. using stackable-utils): if this is not done then an error ocurs in the workflow e.g. `Cannot downgrade from 24.11.1-rc1 to 0.0.0-pr550`

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -13,7 +13,7 @@ on:
       - trying
       - "renovate/**"
     tags:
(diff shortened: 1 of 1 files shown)


razvan (approved):
lgtm

PR #471, opened 2024-12-05
ci: set PR-suffix on PRs against non-main/release branches

adwk67:
See stackabletech/stackable-utils#95 (files)

This PR:

- ensures that PRs against non-main (release) branches also incorporate the PR-number into the image name, so that images from a not-yet-merged release-candidate PR have a name distinct from the merged result
- e.g. `24.11.1-rc1-pr554` (unmerged PR branch) vs. `24.11.1-rc1` (merged and tagged in the rlease branch)

### DIFF
--- a/template/.github/workflows/build.yml.j2
+++ b/template/.github/workflows/build.yml.j2
@@ -344,10 +344,15 @@ jobs:
         with:
           crate: cargo-edit
           bin: cargo-set-version
-      - name: Update version if PR
-        # For PRs to be merged against a release branch, the version has already been set, in which case ignore this step.
+      - name: Update version if PR against main branch
         if: ${{ github.event_name == 'pull_request' && github.ref == 'refs/heads/main' }}
         run: cargo set-version --offline --workspace 0.0.0-pr${{ github.event.pull_request.number }}
+      - name: Update version if PR against non-main branch
+        if: ${{ github.event_name == 'pull_request' && github.ref != 'refs/heads/main' }}
+        run: |
+          MANIFEST_VERSION=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')
+          PR_VERSION="${MANIFEST_VERSION}-pr${{ github.event.pull_request.number }}"
+          cargo set-version --offline --workspace "$PR_VERSION"
 
       # Recreate charts and publish charts and docker image. The "-e" is needed as we want to override the
       # default value in the makefile if called from this action, but not otherwise (i.e. when called locally).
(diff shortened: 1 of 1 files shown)


Techassi (approved):

template/.github/workflows/build.yml.j2
```diff
@@ -411,10 +417,16 @@ jobs:
...
-        # For PRs to be merged against a release branch, the version has already been set, in which case ignore this step.
+      - name: Update version if PR against main branch
         if: ${{ github.event_name == 'pull_request' && github.ref == 'refs/heads/main' }}
         run: cargo set-version --offline --workspace 0.0.0-pr${{ github.event.pull_request.number }}
+      - name: Update version if PR against non-main branch
+        # For PRs to be merged against a release branch, use the version that has already been set in the calling script.
+        if: ${{ github.event_name == 'pull_request' && github.ref != 'refs/heads/main' }}
+        run: |
+          MANIFEST_VERSION=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')
+          PR_VERSION="${MANIFEST_VERSION}-pr${{ github.event.pull_request.number }}"
```
NickLarsenNZ, line 428: Ideally we shouldn't use github interpolation in scripts, but rather create env vars for the step/job/workflow, and use those in the script.

In this case, it is ok, since the value can only be a number, and no injection can occur, and the workflow will be restructured to work more like the docker-images workflows.

template/.github/workflows/build.yml.j2
```diff
@@ -411,10 +417,16 @@ jobs:
...
+          cargo set-version --offline --workspace "$PR_VERSION"
```
NickLarsenNZ, line 429: Ideally we shouldn't use github interpolation in scripts, but rather create env vars for the step/job/workflow, and use those in the script.

```suggestion
      - name: Update version if PR against non-main branch
        # For PRs to be merged against a release branch, use the version that has already been set in the calling script.
        if: ${{ github.event_name == 'pull_request' && github.ref != 'refs/heads/main' }}
        env:
          PR_NUMBER: ${{ github.event.pull_request.number }}
        run: |
          MANIFEST_VERSION=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')
          PR_VERSION="${MANIFEST_VERSION}-pr${PR_NUMBER}"
          cargo set-version --offline --workspace "$PR_VERSION"
```

In this case, it is ok, since the value can only be a number, and no injection can occur, and the workflow will be restructured to work more like the docker-images workflows.

NickLarsenNZ: I'll make another suggestion for the other occurrence. Feel free to ignore the suggestions as we can clean it up during the restructure.

template/.github/workflows/build.yml.j2
```diff
@@ -344,10 +344,16 @@ jobs:
...
```
NickLarsenNZ, line 356:
```suggestion
      - name: Update version if PR against non-main branch
        # For PRs to be merged against a release branch, use the version that has already been set in the calling script.
        if: ${{ github.event_name == 'pull_request' && github.ref != 'refs/heads/main' }}
        env:
          PR_NUMBER: ${{ github.event.pull_request.number }}
        run: |
          MANIFEST_VERSION=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')
          PR_VERSION="${MANIFEST_VERSION}-pr${PR_NUMBER}"
          cargo set-version --offline --workspace "$PR_VERSION"
```

adwk67:
@NickLarsenNZ Thanks for the suggestions: I've committed them, so we an example of how to do it properly until such time as the workflows are refactored.

NickLarsenNZ (approved):
LGTM

PR #476, opened 2025-01-14
chore: set defaults to oci

adwk67:
### Changes

- defaults (e.g. labels) are set to `oci.stackable.tech`

#### N.B. not changed (yet)

- objects are still pushed to Nexus and Harbor
  - post 25.3, pushes to Nexus will be discontinued
  - see [this](https://github.com/stackabletech/opa-operator/pull/671/commits/47ff0d19dc14623f289bf7a9e5d17ca0da9486b2) commit for necessary changes to remove Nexus actions


xeniape (approved):
LGTM

PR #477, opened 2025-01-15
chore: Add rustfmt.toml with some unstable format options

sbernauer:
Fixup of https://github.com/stackabletech/operator-templating/pull/136
Back in the day I put it in the wrong folder :man_facepalming: We already have this in secret-op for ~2 years now.

The amount of time I have seen suggestions regarding merging imports and sorting (std, external, crate), I'm very happy that we can solve all of that by automation :partying_face: 

`cargo +stable fmt` currently gives a warning, but still returns exit code 0 (so can be ignored)
Use `cargo fmt` to run the formatting.

See https://github.com/stackabletech/operator-templating/pull/136
> For what it's worth, you can run a single cargo command using the nightly toolchain by using `cargo +nightly fmt`.
>
> Rust-Analyzer should also allow you to set rustfmt-specific flags (for Emacs, put `(setq rust-rustfmt-switches '("+nightly" "--edition" "2018"))` into your `.emacs`), and I assume Rust-IntelliJ has something similar.

### DIFF
--- a/template/.pre-commit-config.yaml.j2
+++ b/template/.pre-commit-config.yaml.j2
@@ -18,6 +18,7 @@ repos:
     rev: eeee35a89e69d5772bdee97db1a6a898467b686e # 1.0
     hooks:
       - id: fmt
+        entry: cargo +nightly fmt
         args: ["--all", "--", "--check"]
       - id: clippy
         args: ["--all-targets", "--", "-D", "warnings"]
--- a/template/.pre-commit-config.yaml.j2
+++ b/template/.pre-commit-config.yaml.j2
@@ -18,7 +18,7 @@ repos:
     rev: eeee35a89e69d5772bdee97db1a6a898467b686e # 1.0
     hooks:
       - id: fmt
-        entry: cargo +nightly fmt
+        entry: RUSTUP_TOOLCHAIN=nightly-2025-01-15 cargo fmt
         args: ["--all", "--", "--check"]
       - id: clippy
         args: ["--all-targets", "--", "-D", "warnings"]
--- a/template/.pre-commit-config.yaml.j2
+++ b/template/.pre-commit-config.yaml.j2
@@ -18,6 +18,7 @@ repos:
     rev: eeee35a89e69d5772bdee97db1a6a898467b686e # 1.0
     hooks:
       - id: fmt
+        # Pinning to a specific rustc version, so that we get consistent formatting
         entry: RUSTUP_TOOLCHAIN=nightly-2025-01-15 cargo fmt
         args: ["--all", "--", "--check"]
       - id: clippy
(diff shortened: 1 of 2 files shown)


razvan (approved):

sbernauer:
@siegfriedweber had a good point that this might result in back and forth changes between developer machines with different nightly versions.
We could pin a specific nightly version, but that implies some maintenance burden.

As `kube` is also using a nightly formatter I would like to give it a try and see if that is really the case. In that case we should be able to pin a specific nightly version

template/.pre-commit-config.yaml.j2
```diff
@@ -18,6 +18,7 @@ repos:
     rev: eeee35a89e69d5772bdee97db1a6a898467b686e # 1.0
     hooks:
       - id: fmt
+        entry: cargo +nightly fmt
```
Techassi, line 21: question: Is overriding the entry key supported in non-local hooks? Because https://github.com/doublify/pre-commit-rust/issues/15 seems to suggest it is not.

If this is the case, we should consider using a local hook instead.

Techassi (changes requested):

nightkr:
Pinning sounds good to me. It's not like we'd need to bump this more often than we already do for rustc anyway.

template/.pre-commit-config.yaml.j2
sbernauer, line 21: Is this what you meant @nightkr?

It has the downside that users need to understand and install this
`error: toolchain 'nightly-2025-01-15-x86_64-unknown-linux-gnu' is not installed`
This also applies to CI and such

I'm not sure if this complexity is worth it, seeing kube only using `+nightly`

```suggestion
        entry: cargo +nightly-2025-01-15 fmt
```

nightkr: Looks like `RUSTUP_TOOLCHAIN=beta cargo fmt` _will_ automatically install it if missing.

nightkr: image

sbernauer: Can confirm, `RUSTUP_TOOLCHAIN=nightly-2025-01-15 cargo fmt` works for me :partying_face: 

Should we use that for pre-commit?
So `entry: RUSTUP_TOOLCHAIN=nightly-2025-01-15 cargo fmt`

nightkr: I'd say go for it.

sbernauer: I personally have to admit I don't know, I don't use pre-commit :see_no_evil:

razvan: +1 for replacing pre-commit-rust altogether

sbernauer:
We decided to go with this in a daily last week. If there are any problems we can always revert this change

Techassi (approved):

PR #480, opened 2025-01-30
fix(template/helm): Add checksum/config annotation

NickLarsenNZ:
Add the dynamic `checksum/config` annotation back in.
This is supposed to be used for rolling out the Deployment when the ConfigMap content changes.

Previously, this annotation only appeared when `.Values.podAnnotations` was set.

It was only discovered broken by adding podAnnotations and seeing that the `include` context was invalid, leading to this error:

```
Error: template: trino-operator/templates/deployment.yaml:20:28: executing "trino-operator/templates/deployment.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: error calling include: template: trino-operator/templates/configmap.yaml:4:10: executing "trino-operator/templates/configmap.yaml" at <.Files.Glob>: nil pointer evaluating interface {}.Glob
```

Note, this was always broken since inception in #7.

### DIFF
--- a/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
+++ b/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
@@ -17,8 +17,8 @@ spec:
     metadata:
       annotations:
         internal.stackable.tech/image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
-        {{- with .Values.podAnnotations }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
       labels:


NickLarsenNZ:
# Release Note

Operators deployed by Helm will contain an extra pod annotation: `checksum/config`.

NickLarsenNZ:
# Release Note

Operators deployed by Helm will contain an extra pod annotation: `checksum/config`.

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
```diff
@@ -17,8 +17,8 @@ spec:
     metadata:
       annotations:
         internal.stackable.tech/image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
-        {{- with .Values.podAnnotations }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
Comment (NickLarsenNZ) [inline template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:20]: Should we add a qualifier to the annotation?

Techassi (approved):
Approved as is.

I don't have any strong opinions on namespacing the checksum field.

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
Comment (NickLarsenNZ) [reply template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:20]: It appears there are no strong feelings here. I will leave it as is. It can be changed in future if need be.

PR #483, opened 2025-02-04
fix(template/ci): Use local rustfmt hook

Techassi:
This fixes the rustfmt pre-commit hook introduced in #477. Pre-commit tries to run `RUSTUP_TOOLCHAIN=nightly-2025-01-15` instead of `cargo fmt`, see [CI run](https://github.com/stackabletech/druid-operator/actions/runs/13131269765/job/36636760099#step:4:524).


NickLarsenNZ (approved):
LGTM

PR #485, opened 2025-03-03
chore: Pre-release updates for Stackable Release 25.3.0

Techassi:
Part of <https://github.com/stackabletech/operator-templating/issues/484>

- Bumps local and templated workflow actions
- Hardens local and templated workflows
- Bumps local and templated pre-commit hooks
- Bumps the Rust toolchain to 1.84.1

Issue #484, opened 2025-03-03
chore(tracking): Pre-release updates for Stackable Release 25.3.0


Techassi:
## Pre-Release Updates for Stackable Release 25.3.0

Part of <https://github.com/stackabletech/issues/686>

### Update pre-commit Workflow

> [!NOTE]
> The pre-commit config and workflows need to be kept up-to-date to ensure usage of recent tooling versions. This requires some manual work in this repository.

```[tasklist]
### Tasks
- [x] ~Update `python-version` in local and templated `pr_pre-commit.yml` workflow~
- [x] Update hook refs in local and templated `.pre-commit-config.yaml` file
- [x] ~Update Hadolint version in the `config/versions.yaml` file~
```

### Update Rust Toolchain

> [!NOTE]
> During a Stackable release we need to ensure that every product operator uses the latest Rust toolchain (used by us). To keep the toolchain in sync across all our operators, we update the version centrally in this repository.

```[tasklist]
### Tasks in this Repository
- [x] Update Rust toolchain in the `config/versions.yaml` file.
- [x] Update Rust toolchain in ubi8 and ubi9 images
- [x] Generate downstream PRs using the ["Generate Downstream PRs"](https://github.com/stackabletech/operator-templating/actions/workflows/generate_prs.yml) action.
- [x] [Search for PRs](https://github.com/search?q=org%3Astackabletech%20sort%3Aupdated-desc%20is%3Apr%20is%3Aopen%20Update%20templated%20files&type=pullrequests) and add them to the task list below.
- [x] Merge downstream PRs, see below for more details.
```

### Tasks

```[tasklist]
### Related tasks
- [ ] https://github.com/stackabletech/operator-templating/pull/485
- [ ] https://github.com/stackabletech/docker-images/pull/1012
- [ ] https://github.com/stackabletech/docker-images/pull/1013
- [ ] https://github.com/stackabletech/operator-templating/pull/486
```

Replace the items in the task lists below with the applicable Pull Requests

```[tasklist]
### Tasks in Downstream Operator Repositories
- [ ] https://github.com/stackabletech/airflow-operator/pull/590
- [ ] https://github.com/stackabletech/commons-operator/pull/333
- [ ] https://github.com/stackabletech/druid-operator/pull/694
- [ ] https://github.com/stackabletech/edc-operator/pull/155
- [ ] https://github.com/stackabletech/hbase-operator/pull/633
[...]

PR #486, opened 2025-03-04
template(ci): Bump cargo-deny-action to 2.0.7

Techassi:
Part of <https://github.com/stackabletech/operator-templating/issues/484>

A new version of the cargo-deny-action was released because of the rustup 1.28 change of behaviour, see https://github.com/EmbarkStudios/cargo-deny-action/pull/92.

This change is already rolled out using the PRs from the previous rollout of #485.


NickLarsenNZ (approved):
LGTM

PR #489, opened 2025-03-19
Add functionality to allow configuring the loglevel the operator is deployed with in the Helm chart

soenkeliebau:
This introduces a variable called 'logLevel' which, when set to an empty string is not propagated to the deployment by the Helm chart at all, so that the default value for the loglevel in the operator code remains effective.

When set to a non-empty string the value is propagated as environment variable <PRODUCTNAME>_OPERATOR_LOG into the pod template of the operator deployment.

Additionally this adds code to the Tiltfile that allows configuring this via a local override, in a similar fashion to what we already allow for the default registry.

Users can create a file called `tilt_options.json` and in there define a loglevel as follows: 
```json
{
  "loglevel": "DEBUG"
}
```

When this file is present the value from here will be read and passed to the Helm chart by Tilt. If this file is not present, an empty string is passed through, triggering the behavior described above where the default loglevel defined in the operator code remains effective.

This PR doesn't introduce a default for the 'logLevel' variable in the Helm charts, as the `values.yaml` files are not managed by templating. However adding a default for that variable should be an optional step, that we can do later at our convenience, or totally omit, since an unset value will just fall through to the default loglevel from code again.

### DIFF
--- a/template/Tiltfile
+++ b/template/Tiltfile
@@ -1,6 +1,7 @@
 # If tilt_options.json exists read it and load the default_registry value from it
 settings = read_json('tilt_options.json', default={})
 registry = settings.get('default_registry', 'oci.stackable.tech/sandbox')
+loglevel = settings.get('loglevel', '')
 
 # Configure default registry either read from config file above, or with default value of "oci.stackable.tech/sandbox"
 default_registry(registry)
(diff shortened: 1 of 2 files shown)


adwk67:
I'm fine with this, as I asked the question that triggered it and have tested it locally, but would be happy to let others approve (or not).

soenkeliebau:
I'd say, lets definitely wait until release branches have been cut with this.
While in theory not critical, it does touch the helm charts and hence has the ability to break everything.

PR #494, opened 2025-03-24
Update issue templates

lfrancke:


template/.github/ISSUE_TEMPLATE/02-bug_report.yml.j2
NickLarsenNZ, line 1: SInce the file has been touched, it needs to pass linting.

I believe the change needs to look something like this:

```yaml
# {[% if operator.has_product | default(true) %}]
  - type: input
    attributes:
      label: Affected {[ operator.pretty_string }] version
      description: Which version of {[ operator.pretty_string }] do you see this bug in?
# {[% endif %}]
```

NickLarsenNZ: Done in 554415035acc0db0fee24cae6b92b65882944c22

NickLarsenNZ (approved):
LGTM

PR #495, opened 2025-04-02
chore(template): Bump Rust to 1.85.0

NickLarsenNZ:
- Add additional rustfmt settings used in operator-rs
- Use Rust 2024 style
- Add vscode settings for rustfmt
- Bump rust to 1.85.0


template/.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "rust-analyzer.rustfmt.overrideCommand": [
+        "rustfmt",
+        "+nightly-2025-01-15",
```
Techassi, line 4: note: We should use the `rust_nightly_version` variable here.

```suggestion
        "{[rust_nightly_version}]",
```

This also means that this file needs to be renamed to `settings.json.j2`

Techassi (changes requested):
LGTM, just one requested change.

PR #497, opened 2025-04-02
fix(template): Use nightly toolchain for cargo udeps

NickLarsenNZ:
> [!CAUTION]
> Do not merge. It hasn't seemed to fix the problem in CI (though it does work locally).

`cargo udeps` was failing on the `1.85.0` toolchain, but it seems to work on nightly.


NickLarsenNZ:
Wasn't needed. A new cargo-udeps was built: https://github.com/stackabletech/cargo-install-action

PR #498, opened 2025-04-03
chore(template): Template Rust edition in rustfmt config

Techassi:
If the `rustfmt` command in `.vscode/settings.json` is overwritten, the edition needs to be provided explicitly as otherwise the command will fail and no formatting on save is performed.


NickLarsenNZ (approved):
LGTM

PR #502, opened 2025-04-11
chore: Remove Nexus jobs/workflows

xeniape:
Part of https://github.com/stackabletech/infrastructure/issues/142 After Release Tasks
This PR removes jobs and workflows building for and pushing to nexus


template/.github/workflows/build.yml.j2
```diff
@@ -397,21 +384,21 @@ jobs:
...
 
-          make -e build
+          make build
       - name: Publish Docker image and Helm chart
         if: ${{ !github.event.pull_request.head.repo.fork }}
         run: |
           # We want to publish helmcharts only once as they have a common name, while still publishing both images with architecture specific tags
           if [ "$(uname -m)" = "x86_64" ]; then
-            make -e publish
+            make publish
```
NickLarsenNZ, line 393: Just checking that removing the `-e` is intentional and well thought out here.
I believe that removing them might prevent people being able to override things via the CLI, eg:

```shell
SOME_VAR=overridden make blah # if SOME_VAR is set to a default in the Makefile
```

NickLarsenNZ (approved):
LGTM, just left a comment about the removal of `-e`.

template/.github/workflows/build.yml.j2
xeniape: As far as I understood - and the removed comment above suggested as much - it was only needed for the `HELM_REPO` variable, which had a default in the Makefile but we also set it as an env here depending on the Helm repo being used. Now we are only setting `OCI_REGISTRY_SDP_PASSWORD` , `OCI_REGISTRY_SDP_USERNAME`, `OCI_REGISTRY_SDP_CHARTS_PASSWORD`, and `OCI_REGISTRY_SDP_CHARTS_USERNAME` which all don't have a default value in the Makefile, so we don't need that flag anymore. But don't know if that counts as "well thought out", feel free to correct me if my assumptions are wrong :smile:

NickLarsenNZ: That seems well thought out to me, so no problem.

PR #501, opened 2025-04-10
chore(template): Add Helm helper for telemetry env vars

Techassi:
Part of https://github.com/stackabletech/issues/issues/639

Depends on
- https://github.com/stackabletech/operator-rs/pull/1009
- https://github.com/stackabletech/operator-rs/pull/1010
- https://github.com/stackabletech/operator-rs/pull/1012

This adds the `telemetry.envVars` Helm helper to create a list of environment variables passed on the `.telemetry` values.

**Example**

Given these values form the Helm `values.yaml` file

```yaml
telemetry:
  consoleLog:
    enabled: false
    level: warn
    format: json
  fileLog:
    enabled: true
    level: debug
    rotationPeriod: hourly
    maxFiles: 6
  otelLogExporter:
    enabled: true
    level: trace
    endpoint: my-collector:4317
  otelTraceExporter:
    enabled: true
    level: error
    endpoint: my-collector:4317
```

the rendered env block in the Deployment looks like

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo-operator-airflow-operator-deployment
spec:
  template:
    spec:
      containers:
        - name: airflow-operator
          env:
            - name: CONSOLE_LOG_DISABLED
              value: "true"
            - name: CONSOLE_LOG_LEVEL
              value: warn
            - name: CONSOLE_LOG_FORMAT
              value: json
            - name: FILE_LOG_DIRECTORY
              value: /stackable/logs/airflow-operator
            - name: FILE_LOG_LEVEL
              value: debug
            - name: FILE_LOG_ROTATION_PERIOD
              value: hourly
            - name: FILE_LOG_MAX_FILES
              value: 6
            - name: OTEL_LOG_EXPORTER_ENABLED
              value: "true"
            - name: OTEL_LOG_EXPORTER_LEVEL
              value: trace
            - name: OTEL_EXPORTER_OTLP_LOGS_ENDPOINT
              value: my-collector:4317
            - name: OTEL_TRACE_EXPORTER_ENABLED
              value: "true"
            - name: OTEL_TRACE_EXPORTER_LEVEL
              value: error
            - name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
              value: my-collector:4317
```

Default values will be set according to the design laid out in the [tracking issue](https://github.com/stackabletech/issues/issues/639).

PR #505, opened 2025-04-15
chore(template): Only emit env vars if telemetry signals are enabled

Techassi:
Part of https://github.com/stackabletech/issues/issues/639, follow-up of #501 

This improves the telemetry Helm helper to only emit env vars (e.g. for level) when the appropriate signal is enabled. This will no longer emit env vars even though the signal it self is disabled.


NickLarsenNZ (approved):
LGTM

PR #508, opened 2025-04-16
ci(generate_prs): Set fail-fast input default value to false

Techassi:
This is what we want in most situations, so set the default accordingly.


NickLarsenNZ (approved):
LGTM

PR #513, opened 2025-05-02
chore(template): Remove cargo deny exclusions

NickLarsenNZ:
Part of https://github.com/stackabletech/issues/issues/677

> [!NOTE]
> I did consider leaving one excluding in, and commenting it so that we have a template for future exclusions. Let me know if you prefer that.


razvan (approved):
lgtm

PR #516, opened 2025-05-07
ci(template): Install Nix in pre-commit workflow

Techassi:
Came up in https://github.com/stackabletech/secret-operator/actions/runs/14877437184/job/41777443726.

Currently, the pre-commit workflow will fail, because Nix (the package manager) is not installed. This PR installs it in the latest version (2.28.3).

### DIFF
--- a/config/versions.yaml
+++ b/config/versions.yaml
@@ -23,3 +23,7 @@ hadolint_version: v2.12.0
 # If you change the Python version here, make sure to also change it in
 # .github/workflows/pr_pre-commit.yml
 python_version: 3.12
+
+# This Nix version is used in the operators pre-commit workflow as they require
+# Nix to ron some checks.
+nix_pkg_manager_version: 2.28.3
--- a/config/versions.yaml
+++ b/config/versions.yaml
@@ -25,5 +25,5 @@ hadolint_version: v2.12.0
 python_version: 3.12
 
 # This Nix version is used in the operators pre-commit workflow as they require
-# Nix to ron some checks.
+# Nix to run some checks.
 nix_pkg_manager_version: 2.28.3
--- a/config/versions.yaml
+++ b/config/versions.yaml
@@ -25,9 +25,8 @@ hadolint_version: v2.12.0
 python_version: 3.12
 
 # This Nix version is used in the operators pre-commit workflow as they require
-# Nix to run some checks. Places to check for versions:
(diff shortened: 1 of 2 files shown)


config/versions.yaml
```diff
@@ -23,3 +23,7 @@ hadolint_version: v2.12.0
 # If you change the Python version here, make sure to also change it in
 # .github/workflows/pr_pre-commit.yml
 python_version: 3.12
+
+# This Nix version is used in the operators pre-commit workflow as they require
+# Nix to ron some checks.
```
NickLarsenNZ, line 28:
```suggestion
# Nix to run some checks.
```

config/versions.yaml
```diff
@@ -23,3 +23,7 @@ hadolint_version: v2.12.0
 # If you change the Python version here, make sure to also change it in
 # .github/workflows/pr_pre-commit.yml
 python_version: 3.12
+
+# This Nix version is used in the operators pre-commit workflow as they require
+# Nix to run some checks.
```
NickLarsenNZ, line 28: Could be useful to add a comment where to check versions (or mention any constraints before upgrading).

```suggestion
# This Nix version is used in the operators pre-commit workflow as they require
# Nix to run some checks.
# Check for new versions here: https://nixos.org/download/
```

Ideally we should do this for all version references (both where to check, and upgrade constraints).

Techassi: Agreed, that makes sense. Let me add a link regarding the version reference.

Techassi: Done on 8a6d32b4cda3f95899161b1ef86b14c925a020d9.

config/versions.yaml
```diff
@@ -23,3 +23,11 @@ hadolint_version: v2.12.0
 # If you change the Python version here, make sure to also change it in
 # .github/workflows/pr_pre-commit.yml
 python_version: 3.12
+
+# This Nix version is used in the operators pre-commit workflow as they require
+# Nix to run some checks. Places to check for versions:
+# - https://github.com/NixOS/nix/tags
+# - https://nixos.org/download/#
```
NickLarsenNZ, line 30:
```suggestion
# This Nix version is used in the operators pre-commit workflow as they require
# Nix to run some checks. Check for new versions here:
# https://github.com/NixOS/nix/tags
```

NickLarsenNZ (approved):
LGTM

PR #520, opened 2025-05-16
Adding sed compatibility for MacOS

Maleware:
This PR adds:

- compatibility increase for MacOS for NixOS powered dev tools


NickLarsenNZ (approved):
LGTM

nightkr:
I'd rather just force it to use a consistent sed, whether that's bsdsed or gnused. Your kernel is not a reliable predictor for your sed.

PR #525, opened 2025-05-31
Hardcode the use of the "default" profile for rustup

lfrancke:
At some point rustup changed the default profile from "default" to "minimal" if no other profile was specified.

This caused our pre-commit checks to fail which rely on clippy which is not in the "minimal" profile.
We now specify our dependency on the "default" profile explicitly to be more independent of future rustup changes.


lfrancke:
I created an example PR here: https://github.com/stackabletech/hive-operator/pull/606

Cargo will automatically install missing components/profiles if they are not already installed but specified in the `rust-toolchain.toml` file.

nightkr (approved):
Alternatively, we might want to list out the specific components we care about. But this is fine for now.

nightkr:
FWIW, this will end up in the same templating batch as #524, so we'll likely need to manually `cargo +nightly fmt --all` each PR.

lfrancke:
Yeah, my initial attempt was to list all components but the docs say they can change between versions of rust.
I have no strong opinion here. If anyone prefers I can also put up a PR to list the current components from `default` profile.

This will also include the OpenSearch operator inclusion which might also cause this to fail because that repo is private at the moment. If it does we'll have to remove it again for now.

PR #530, opened 2025-06-10
Slim down Nix builds

nightkr:
Primarily by disabling debug info.

For example, this slims down the airflow-operator image from 615MiB to 147MiB, and takes the incremental redeploy time from 57s to 29s (on my machine).

This improvement is a combination of the size of the debuginfo itself (~200MiB per binary), and not having to include build dependencies just because our debuginfo refers to them (primarily, this ended up pulling in GCC, which sits at ~260MiB just on its own).

The old behaviour (useful if you want to run it under a debugger.. though we probably want to support that separately somehow) can be restored by adding `--arg debuginfo 2` to the `nix-build` call.

In addition, this PR adds an option to remove the shell (and associated utililties) from the built images. This can be done by setting `--arg includeShell false`, but is disabled by default because not all operators are compatible yet (in particular: https://github.com/stackabletech/secret-operator/issues/608).

### DIFF
--- a/template/default.nix
+++ b/template/default.nix
@@ -36,45 +36,84 @@
     # We're only using this for dev builds at the moment,
     # so don't pay for release optimization.
     release = false;
-    defaultCrateOverrides = pkgs.defaultCrateOverrides // {
-      prost-build = attrs: {
-        buildInputs = [ pkgs.protobuf ];
+
+    buildRustCrateForPkgs = pkgs: attrs: pkgs.buildRustCrate.override {
+      # Consider migrating to mold for faster linking, but in my (@nightkr's)
+      # quick testing so far it actually seems to perform slightly worse than
+      # the default one.
+      # stdenv = pkgs.stdenvAdapters.useMoldLinker pkgs.stdenv;
+
+      defaultCrateOverrides = pkgs.defaultCrateOverrides // {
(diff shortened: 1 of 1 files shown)


template/default.nix
```diff
@@ -36,45 +36,90 @@
...
+# see https://doc.rust-lang.org/cargo/reference/profiles.html#debug
+# Set to 2 if you want to run a debugger, but note that it bloats the Docker
+# images *significantly* (hundreds of megabytes).
+, debuginfo ? 0
+# Strip operator binaries if we don't include debuginfo, because *something*
+# still something still includes a reference to gcc (~230MiB), causing it to be
+# added to the docker images.
+, strip ? if debuginfo == 0 then true else false
+# We normally don't include a shell in the (dev) operator images, but it can be
+# enabled by enabling this flag.
```
NickLarsenNZ, line 118: A shell can be super handy in dev, so I'd be inclined to keep it in by default and allow it to be removed if there is some performance gain.

NickLarsenNZ (approved):
LGTM, tested with nifi-operator.

It did still take a little long, (`DONE IN: 102.86s`) but I guess that's just my machine, and maybe the operator I chose to test on.

nightkr:
Yeah that's still pretty slow, but also not super insightful without a before-number to compare against.

PR #531, opened 2025-06-11
feat(run-tests): Handle existing namespaces and permission errors

lfrancke:
Handle existing namespaces and permission errors more gracefully

This _does_ change behavior.
In the past run-tests would fail if we can't create the namespace but now we'll proceed because it might already exist and we just can't check easily.


NickLarsenNZ (approved):
I like the idea. I have needed this.

Too bad that Helm usually kills the fun (when charts already exist).

PR #532, opened 2025-06-11
Restore repository indexing

### LINKED ISSUE #411: Dynamic port number for tilt not working any more
When we parallelized the PR generation in https://github.com/stackabletech/operator-templating/commit/1ce22e88a5c3d7c3adef01ef080ec9697202a9c1 this broke the [generation](https://github.com/stackabletech/operator-templating/blob/main/template/Makefile.j2#L193) of a unique port number for Tilt. When run from github actions the loop index will always be 0 because the workflow replaces the repository list with a list with only one entry - so the port will be the same for every repository.

This is not a huge problem I guess, can always be edited manually .. but if we decide to go that route we should remove the incrementation code linked above because it'll just cause confusion.

Alternatively we can probably come up with a different way of doing this, if we want to keep the functionality. Keeping the port number in the metaqata would work, but require us to track it .. 

In principle I have to say that I'm not in love with keeping the repository data in two places, the workflow file and the repositories config file.. maybe we could somehow read the matrix data for the workflow from the yaml file? 

https://stackoverflow.com/questions/63517732/github-actions-build-matrix-for-lambda-functions/63736071#63736071 seems to suggest this might be possible..

nightkr:
Fixes #411.

config/repositories.yaml is now authoritative (once again) for the details of repository configuration (but generate_prs.yml must still contain the _names_ of all operators). Ansible now knows about the whole list of repositories and is responsible for filtering out the ones it cares about, rather than GitHub Actions generating a unique repository list for every shard. This means that Ansible is now able to generate proper indexes again.

Also refactors the github workflow a bit to avoid different code paths for dry and wet runs.

### DIFF
--- a/.github/workflows/generate_prs.yml
+++ b/.github/workflows/generate_prs.yml
@@ -113,10 +113,10 @@ jobs:
         env:
           DEBIAN_FRONTEND: noninteractive
         run: |
-            sudo apt update &&  \
-            sudo apt install -y software-properties-common && \
-            sudo apt-add-repository ppa:ansible/ansible -y && \
-            sudo apt install -y ansible
+          sudo apt update &&  \
+          sudo apt install -y software-properties-common && \
+          sudo apt-add-repository ppa:ansible/ansible -y && \
+          sudo apt install -y ansible
 
       # NOTE (@NickLarsenNZ): This could be removed in favor of nix-shell and rrbutani/use-nix-shell-action
       - name: Install deps for operators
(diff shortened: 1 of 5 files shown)


NickLarsenNZ (approved):
LGTM

Has the workflow been tested?

nightkr:
I've tested the port numbering locally, and confirmed with a [dry run](https://github.com/stackabletech/operator-templating/actions/runs/15582908746/job/43882175804) that the sharding and, well, dry-run functionality still seems to work. Not sure why only airflow-operator is listed as having been modified in the CI logs...

PR #537, opened 2025-07-09
feat: Add KUBERNETES_NODE_NAME env var to operator Pods

sbernauer:
Needed by https://github.com/stackabletech/operator-rs/pull/1071

### DIFF
--- a/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
+++ b/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
@@ -44,19 +44,25 @@ spec:
               name: config-spec
           env:
             # The following env vars are passed as clap (think CLI) arguments to the operator.
-            # They are picked up by clap using the structs defied in
-            # https://github.com/stackabletech/operator-rs/blob/main/crates/stackable-operator/src/cli.rs
-            # (currently "ProductOperatorRun").
+            # They are picked up by clap using the structs defied in the operator.
+            # (which is turn pulls in https://github.com/stackabletech/operator-rs/blob/main/crates/stackable-operator/src/cli.rs)
             # You can read there about the expected values and purposes.
+
+            # Sometimes products need to know the operator image, e.g. the opa-bundle-builder OPA
+            # sidecar uses the operator image.
             - name: OPERATOR_IMAGE
               # Tilt can use annotations as image paths, but not env variables
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.annotations['internal.stackable.tech/image']
+
+            # Operators need to know the node name they are running on, to e.g. discover the
+            # Kubernetes domain name from the kubelet API.
             - name: KUBERNETES_NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+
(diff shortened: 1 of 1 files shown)


razvan (approved):
I would add a small comment explaining why it's needed.

sbernauer:
I added a small explanation what the env vars are and how to find about the meaning in 52047c1
I would like to avoid doubling information, does that work for you?

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
```diff
@@ -43,11 +43,20 @@ spec:
...
+            # They are picked up by clap using the structs defied in
+            # https://github.com/stackabletech/operator-rs/blob/main/crates/stackable-operator/src/cli.rs
+            # (currently "ProductOperatorRun").
+            # You can read there about the expected values and purposes.
             - name: OPERATOR_IMAGE
               # Tilt can use annotations as image paths, but not env variables
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.annotations['internal.stackable.tech/image']
+            - name: KUBERNETES_NODE_NAME
```
Comment (razvan) [inline template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:56]:
```suggestion
            # Used by the operators to discover the Kubernetes domain name from the kubelet API.
            - name: KUBERNETES_NODE_NAME
```

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
```diff
@@ -43,11 +43,20 @@ spec:
             - mountPath: /etc/stackable/{{ include "operator.appname" . }}/config-spec
               name: config-spec
           env:
+            # The following env vars are passed as clap (think CLI) arguments to the operator.
+            # They are picked up by clap using the structs defied in
+            # https://github.com/stackabletech/operator-rs/blob/main/crates/stackable-operator/src/cli.rs
+            # (currently "ProductOperatorRun").
+            # You can read there about the expected values and purposes.
```
Comment (razvan) [inline template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:50]: IMO referencing to some obscure Rust code is no good. Maybe I'm blind, but I didn't find what `OPERATOR_IMAGE` is actually needed for and I really don't know.

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
Comment (sbernauer) [reply template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:61]: I tried to avoid hardcoding the purpose here. Currently it's only used for that, but it might get outdated in the past.
I changed it to be non-exhausting
```
# Operators need to know the node name they are running on, to e.g. discover the
# Kubernetes domain name from the kubelet API.
```
Is that ok?

template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
Comment (sbernauer) [reply template/deploy/helm/[[operator]]/templates/deployment.yaml.j2:49]: https://github.com/stackabletech/opa-operator/blob/a5a418c585de8a8ee7efa18de5cf243431cac872/rust/operator-binary/src/main.rs#L50-L52

I updated the comment that the struct is actuality defined in the operator, with 90% of the envs being defined in op-rs. The previous comment was a bit misleading.
I also added a description to OPERATOR_IMAGE in e6771ad

PR #546, opened 2025-08-19
fix(template): Fix whitespace in bug report issue template

Techassi:
Follow-up of #543.

### DIFF
--- a/config/retired_files.yaml
+++ b/config/retired_files.yaml
@@ -3,5 +3,6 @@
 # This is uncommented as I had issues with everything being deleted when this was just present as an empty key.
 # May be something to investigate.
 retired_files:
+  - .github/ISSUE_TEMPLATE/bug_report.yml
   - .readme/static/sdp_overview.png
   - bors.toml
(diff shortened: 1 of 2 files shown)


config/retired_files.yaml
```diff
@@ -3,5 +3,6 @@
 # This is uncommented as I had issues with everything being deleted when this was just present as an empty key.
 # May be something to investigate.
 retired_files:
+  - .github/ISSUE_TEMPLATE/bug_report.yml
```
dervoeti, line 6: Why was it added to the retired files?

Techassi: Because it wasn't added to the retired files in #494 but should have been added to it.

Techassi:
We noticed that this still doesn't completely solve the issues. The comments will need to be indented properly. I will update the PR asap.

dervoeti (approved):

NickLarsenNZ (dismissed):
Requesting that these go into the PR:
* https://github.com/stackabletech/airflow-operator/pull/674#discussion_r2284794002
* https://github.com/stackabletech/airflow-operator/pull/674#discussion_r2285052178

PR #547, opened 2025-08-19
chore: Enable strict mode for yamllint

Techassi:
Came up in the rollout of #544 (and other recent changes), eg. in https://github.com/stackabletech/airflow-operator/pull/674.

### DIFF
--- a/template/.pre-commit-config.yaml.j2
+++ b/template/.pre-commit-config.yaml.j2
@@ -18,6 +18,7 @@ repos:
     rev: 79a6b2b1392eaf49cdd32ac4f14be1a809bbd8f7 # 1.37.1
     hooks:
       - id: yamllint
+        args: ["--strict"]
 
   - repo: https://github.com/igorshubovych/markdownlint-cli
     rev: 192ad822316c3a22fb3d3cc8aa6eafa0b8488360 # 0.45.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,6 +18,7 @@ repos:
     rev: 79a6b2b1392eaf49cdd32ac4f14be1a809bbd8f7 # 1.37.1
     hooks:
       - id: yamllint
+        args: ["--strict"]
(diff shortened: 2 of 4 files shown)


NickLarsenNZ (approved):
LGTM, thanks.

It would pay to run pre-commit across a repo in the `work/` directory if you haven't - just to avoid problems after rolling out.

template/.pre-commit-config.yaml.j2
NickLarsenNZ, line 1: Maybe strict mode should also be enabled in the operator-templating pre-commit config too.

Techassi: That's already done here: https://github.com/stackabletech/operator-templating/blob/c465af0ecf82fb2410fac184a80ec24e4f457982/.pre-commit-config.yaml#L21

Techassi:
> It would pay to run pre-commit across a repo in the work/ directory if you haven't - just to avoid problems after rolling out.

I can do a run, but there will most likely be many errors of non-templated files. I will try to filter those out...

NickLarsenNZ:
> I can do a run, but there will most likely be many errors of non-templated files. I will try to filter those out...

This would be exactly the same as running it on the resultant PR - just earlier. So if there are errors now, then there would be errors when the PRs are raised too.

> A rough looks seems to indicate that almost all the warnings/errors come from non-templated files. I suggest that we fix those in the downstream PRs once this PR is rolled out.

Not sure I understand.

Are you doing:

```sh
# run templaing, then...
cd work/airflow-operator
pre-commit install --install-hooks
pre-commit run --from-ref main --to-ref HEAD
```

Techassi:
> Not sure I understand.

I'm running both the command you suggested, but also without specifying the refs to see what other, non-templated files will be affected. The gist is: This PR looks fine for templated files, so is ready to go.

There however will be fallout for other non-templated files, because there are many YAML files which contain warnings/errors. Those however will only come up if touched.

PR #551, opened 2025-09-16
feat(template/helm): Allow Pod priorityClassName to be configured

NickLarsenNZ:
Part of https://github.com/stackabletech/issues/issues/765.

- Allow Pod `priorityClassName` to be configured.
- Also disable rather useless yamllint lint. It would otherwise stop example values being added as comments.

Note: listener-operator and secret-operator will be excluded from this (since they have `run-mode: custom`), and will be rolled out in their own PRs.

> [!TIP]
> When rolling out, update the changelogs before merging the PRs. Eg:
> ```markdown
> ### Added
>
> - Helm: Allow Pod `priorityClassName` to be configured ([#xxx]).
> ```

### DIFF
--- a/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
+++ b/template/deploy/helm/[[operator]]/templates/deployment.yaml.j2
@@ -104,14 +104,8 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.priority }}
-      priority: {{ . }}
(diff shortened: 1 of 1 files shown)


NickLarsenNZ:
Tested (rendered `nifi-operator`). Found out the admission controller will not allow you to set `priority` and `preemptionPolicy` directly on a Pod. You must use a PriorityClass.

I have removed these.

Techassi (approved):

PR #552, opened 2025-10-01
ci(template): Add hint for the runner or profile to use

NickLarsenNZ:
I never know what to put here, and I imagine others will also be stumped.

_Disclaimer, I happen to know to look in test/interu.yaml (although, I couldn't remember where that was initially)._

image

I think this change makes things more intuitive (at the expense of some ugly UI layout):

image

### DIFF
--- a/template/.github/workflows/integration-test.yml
+++ b/template/.github/workflows/integration-test.yml
@@ -16,7 +16,7 @@ on:
           - custom
       test-mode-input:
         description: |
-          The profile or the runner used. Eg: `amd64` or `smoke-latest` (see test/interu.yaml)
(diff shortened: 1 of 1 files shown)


template/.github/workflows/integration-test.yml
```diff
@@ -15,7 +15,8 @@ on:
           - profile
           - custom
       test-mode-input:
-        description: The profile or the runner used
+        description: |
+          The profile or the runner used. Eg: `amd64` or `smoke-latest` (see test/interu.yaml)
```
Techassi, line 19: suggestion: I would switch the order of the examples to be the same as in the first sentence.

```suggestion
          The profile or the runner used. Eg: `smoke-latest` or `amd64` (see test/interu.yaml)
```

Techassi (approved):

PR #560, opened 2025-11-11
Added option to allow the user the option of not using crate2nix but the default Docker build process instead.

soenkeliebau:
Not sure if this is feasible to use with Tilt due to fairly long build times, might be necessary to pause deployment of the resources in Tilt and manually triggering as and when needed.


soenkeliebau:
full disclosure, not tested it yet, just rescued from https://github.com/stackabletech/operator-templating/tree/feat/multiple_builders

PR #574, opened 2026-02-18
chore: Update templated files for conversion webhook

Techassi:
This brings in changes from https://github.com/stackabletech/issues/issues/808 so that there is no drift between templating and downstream operators anymore.


sbernauer (approved):
Yes, finally :)

PR #577, opened 2026-02-24
template(ci): Fix operator version extraction

Techassi:
Previously, we assumed that the first package is always the operator package. This is not true, for example this resulted in a wrong version here: https://github.com/stackabletech/secret-operator/actions/runs/22309511940/attempts/1.

To fix this, we filter/select the package by operator name instead of hard-coding index `0`.


NickLarsenNZ (approved):
Wow, good catch!

PR #581, opened 2026-03-03
chore(template): Use --no-deps in pre-commit hook

Techassi:
This only generates the local documentation (which is the thing we want to ensure is valid) instead of generating the docs for _everything_.


NickLarsenNZ (approved):
LGTM

PR #586, opened 2026-04-15
chore: Ignore RUSTSEC-2026-0097 in deny.toml

siegfriedweber:
Ignore RUSTSEC-2026-0097 when calling `cargo deny check advisories`

I have not checked if this is an exploitable vulnerability.


razvan (approved):
thanks

PR #587, opened 2026-04-22
chore(template): Combine keys in registry-specific values file

Techassi:
This change is based on https://github.com/stackabletech/decisions/issues/84.

### DIFF
--- a/template/Tiltfile
+++ b/template/Tiltfile
@@ -2,17 +2,18 @@
 meta = read_json('nix/meta.json')
 operator_name = meta['operator']['name']
 
-# If tilt_options.json exists read it and load the default_registry value from it
+# If tilt_options.json exists read it and load the default_registry and default_repository value from it
 settings = read_json('tilt_options.json', default={})
 registry = settings.get('default_registry', 'oci.stackable.tech')
-repository = settings.get('default_repository', 'sandbox' + '/' + operator_name)
+repository = settings.get('default_repository', registry + '/' + 'sdp')
+operator_image_name = repository + '/' + operator_name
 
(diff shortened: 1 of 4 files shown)


NickLarsenNZ (dismissed):
I assume this breaks things until the operator Helm templates are corrected (in the same PR that this generates?)

Techassi:
Good catch, I will update this PR.

NickLarsenNZ (approved):
LGTM

There could be an operator that bypasses the helper and still refers to the image, so we will need to be careful before merging the generated PRs

template/Tiltfile
```diff
@@ -2,17 +2,18 @@
 meta = read_json('nix/meta.json')
 operator_name = meta['operator']['name']
 
-# If tilt_options.json exists read it and load the default_registry value from it
+# If tilt_options.json exists read it and load the default_registry and default_repository value from it
 settings = read_json('tilt_options.json', default={})
 registry = settings.get('default_registry', 'oci.stackable.tech')
-repository = settings.get('default_repository', 'sandbox' + '/' + operator_name)
+repository = settings.get('default_repository', registry + '/' + 'sdp')
```
NickLarsenNZ, line 8: It turns out this breaks local development.

We don't want to push the **operator image** to `sdp`, but instead `sandbox`, but when I change it locally to `sandbox` the the product images fail to pull because they get set to: `oci.stackable.tech/sandbox/kafka:3.9.1-stackable0.0.0-dev` - but I want them to be at `sdp`.

I think we need to separate the concept of the operator image prefix/repo and the product prefix/repo.

PR #590, opened 2026-05-11
feat(template/helm): Split operator+product image repos

Techassi:
Part of https://github.com/stackabletech/issues/issues/716.

This separates the operator from the product image repository (via the Helm values), so that a local `make run-dev` works as before (using `sandbox` for the operator image). Product images are pulled from `sdp` as expected.


NickLarsenNZ (approved):
LGTM

PR #591, opened 2026-05-11
feat(template/helm): Don't include product repo in core operators

Techassi:
Part of https://github.com/stackabletech/issues/issues/716.

### DIFF
--- a/template/Tiltfile.j2
+++ b/template/Tiltfile.j2
@@ -12,9 +12,11 @@ registry = settings.get('default_registry', 'oci.stackable.tech')
 operator_repository = settings.get('default_operator_repository', registry + '/' + 'sandbox')
 operator_image_name = operator_repository + '/' + operator_name
 
+{[% if operator.config.has_product | default(true) %}]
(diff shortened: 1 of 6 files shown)


template/Tiltfile.j2
```diff
@@ -12,9 +12,11 @@ registry = settings.get('default_registry', 'oci.stackable.tech')
 operator_repository = settings.get('default_operator_repository', registry + '/' + 'sandbox')
 operator_image_name = operator_repository + '/' + operator_name
 
+{[% if operator.config.has_product | default(true) %}]
 # For the product image, we wanna use the images in the "sdp" namespace, because "sandbox" doesn't
```
NickLarsenNZ:
```suggestion
# For the product image, we want to use the images in the "sdp" namespace, because "sandbox" doesn't
```

Up to you if you include this or not

NickLarsenNZ (dismissed):
LGTM

I assume it has been tested

NickLarsenNZ (approved):

PR #593, opened 2026-05-26
revert(template): Publish to quay.io (#584)

Techassi:
Part of https://github.com/stackabletech/issues/issues/716.

This reverts commit 96368df17aa9ea9327f18e329bc09d9f53c3ca2b.

This change is reverted because the current default Helm version used by the publish-helm-chart action errors out when trying to push Helm Charts to quay.io. This issue is decribed in the upstream issue: https://github.com/helm/helm/issues/32144.

The only option we currently would have had was to downgrade to Helm 3.17.x. We decided to wait a little longer for either a) a new, fixed Helm version, or b) to downgrade until a fixed version is released.

This revert is needed to unblock other changes in this repo to be rolled out without breaking CI.

PR #584, opened 2026-04-10
ci(template): Publish to quay.io

Techassi:
~~Needs https://github.com/stackabletech/operator-templating/pull/583 to be merged first.~~

Part of https://github.com/stackabletech/issues/issues/716, based on https://github.com/stackabletech/airflow-operator/pull/773.

PR #594, opened 2026-05-26
ci(template): Publish to quay.io

Techassi:
> [!NOTE]
> We downgrade the used Helm version to 3.17.4 (for quay.io) becasue there is no fixed version as of yet.

This change was previously introduced in #584, but reverted in #593.


NickLarsenNZ (approved):
LGTM

PR #596, opened 2026-06-24
ci(template): Bump stackabletech/actions to v0.15.0

Techassi:
This PR bumps all stackabletech/actions to the latest version. The most important change is that many download scripts/actions now do retries when downloads fail.


NickLarsenNZ (approved):
LGTM

PR #608, opened 2026-07-14
Ci hardening

Saul-STFC:
This PR introduces a collection of 'correctness fixes' and 'hygiene' improvements to the CI templates:

Including: 

- Fix failure condition to check actual job results e.g a skipped job will silently pass
- Add missing job dependencies to notify and validate all required jobs in finished
- Guard build-container-image, publish, provenance, preflight and notify jobs with github.repository_owner == 'stackabletech' so fork PRs no longer queue runners or attempt to push images
- Pinned deps to latest hashes
- Some other small hardening improvements

### DIFF
--- a/.github/workflows/generate_prs.yml
+++ b/.github/workflows/generate_prs.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
-      - uses: cachix/install-nix-action@fd24c48048070c1be9acd18c9d369a83f0fe94d7 # v31.8.1
+      - uses: cachix/install-nix-action@630ae543ea3a38a9a4166f03376c02c50f408342 # v31.11.0
       - name: Install Ansible
         env:
           DEBIAN_FRONTEND: noninteractive
@@ -78,15 +78,14 @@ jobs:
       - name: Install jinja2
         run: pip install -r requirements.txt
 
-      # Create commit message and run the playbook
+      # Create commit message depending on whether this is run manually or due to a scheduled run
       - name: Run playbook
         env:
           CUSTOM_PR_TITLE: ${{ inputs.custom-pr-title }}
           GH_ACCESS_TOKEN: ${{ secrets.gh_access_token }}
           REPOSITORY: ${{ matrix.repository }}
-          DRY_RUN_FLAGS: ${{ (github.event_name != 'schedule' && inputs.dry-run != false) && '--tags=local' || '' }}
-          # Safely pass inputs/context via environment variables to prevent injection
-          INPUT_MESSAGE: ${{ github.event.inputs.message }}
+          DRY_RUN_FLAGS: ${{ inputs.dry-run && '--tags=local' || '' }}
+          INPUT_MESSAGE: ${{ inputs.message }}
           SENDER_LOGIN: ${{ github.event.sender.login }}
           EVENT_NAME: ${{ github.event_name }}
         run: |
@@ -98,7 +97,7 @@ jobs:
             REASON="$INPUT_MESSAGE"
           fi
 
-          # Encode parameters safely using jq, without serializing the GH_ACCESS_TOKEN
+          # Encode parameters using jq, without serializing the GH_ACCESS_TOKEN
           jq -n \
             --arg commit_hash "$GITHUB_SHA" \
             --arg author "$AUTHOR" \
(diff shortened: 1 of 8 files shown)


template/.github/workflows/general_daily_security.yml
```diff
@@ -15,8 +15,11 @@ permissions: {}
 jobs:
   audit:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      checks: write
```
Techassi, line 21: I think we should leave a comment here explaining why we need this permission.

.github/workflows/generate_prs.yml
```diff
@@ -77,32 +79,35 @@ jobs:
...
+          EVENT_NAME: ${{ github.event_name }}
         run: |
-          echo '{}' | jq '{commit_hash: $ENV.GITHUB_SHA, author: $ENV.AUTHOR, reason: $ENV.REASON, custom_pr_title: $ENV.CUSTOM_PR_TITLE, base_dir: $pwd, gh_access_token: $ENV.GH_ACCESS_TOKEN, shard_repositories: [$ENV.REPOSITORY]} | with_entries(select(.value != null and .value != ""))' --arg pwd "$(pwd)" > vars.json
-          # $DRY_RUN_FLAGS is intentionally not quoted, since we need to avoid
-          # passing an argument if we're not in dry mode.
-          # This is safe, since it always has one of two hard-coded values.
+          if [ "$EVENT_NAME" = "schedule" ]; then
+            AUTHOR="stackabletech/developers"
+            REASON="Weekly run triggered"
+          else
```
Techassi: We previously explicitly matched on `schedule` XOR `workflow_dispatch`. We should do the same here and fallback to exiting with non-0 if the trigger is neither of the two options.

.github/workflows/generate_prs.yml
```diff
@@ -77,32 +79,35 @@ jobs:
...
+            REASON="$INPUT_MESSAGE"
+          fi
+
+          # Encode parameters using jq, without serializing the GH_ACCESS_TOKEN
+          jq -n \
+            --arg commit_hash "$GITHUB_SHA" \
+            --arg author "$AUTHOR" \
+            --arg reason "$REASON" \
+            --arg custom_pr_title "$CUSTOM_PR_TITLE" \
+            --arg base_dir "$(pwd)" \
```
Techassi: Mhm, curious. This expands to nothing because the correct thing to do here would be `${PWD}`.

template/.github/workflows/build.yaml.j2
```diff
@@ -196,21 +205,23 @@ jobs:
   publish-index-manifest:
     name: Publish/Sign ${{ needs.build-container-image.outputs.operator-version }} Index
     if: |
-      (github.event_name != 'merge_group')
+      github.repository_owner == 'stackabletech'
```
Techassi, line 208: I think this replaces `!github.event.pull_request.head.repo.fork` below, right?

More of these would need to be changed below if that is the case.

Techassi (commented):
Overall looks good to me, thanks!

Just a few minor comments.

template/.github/workflows/build.yaml.j2
Saul-STFC: So github.repository_owner stops a forked repo running the action inside the fork's own copy, whereas !github.event.pull_request.head.repo.fork stops an external PR itself so they are checking for different things.

template/.github/workflows/general_daily_security.yml
Saul-STFC: Done d7b21d0

.github/workflows/generate_prs.yml
Saul-STFC: As this is inside a run block it's just bash

The command below works fine:

jq -n --arg base_dir "$(pwd)" '{base_dir: $base_dir}'

But I can change it to ${PWD} if you want?

.github/workflows/generate_prs.yml
Saul-STFC: Fixed: b482c6f68

.github/workflows/generate_prs.yml
Techassi: I think we should change it to `${PWD}` so that we consistently use env vars.

template/.github/workflows/build.yaml.j2
Techassi: Ah I see. That makes sense, thanks!

.github/workflows/generate_prs.yml
Saul-STFC: Done 5a7a4f9

Techassi (dismissed):

Techassi (dismissed):

Techassi (approved):

PR #612, opened 2026-07-20
ci(template): build cargo-audit against stable rust toolchain

lfrancke:
## Description

The daily security audit has been [failing](https://github.com/stackabletech/hdfs-operator/actions) for a couple of days because a dependency of `cargo-audit` (`kstring`) now requires Rust 1.96.0 to build. See one example
(`general_daily_security.yml`) 

This PR makes it so that cargo-audit builds against the current stable version instead of our pinned one.
I'm fairly certain this fixes the issue itself but the cost will be a slightly longer runtime because it needs to download the second toolchain I guess. Opinions welcome.

Edit: I just realize that runtime doesn't matter much for the nightly cron job...

@StefanFl @dervoeti Not sure if we even still want/need these runs at all?

### DIFF
--- a/template/.github/workflows/general_daily_security.yml
+++ b/template/.github/workflows/general_daily_security.yml
@@ -20,5 +20,10 @@ jobs:
         with:
           persist-credentials: false
       - uses: rustsec/audit-check@69366f33c96575abad1ee0dba8212993eecbe998 # v2.0.0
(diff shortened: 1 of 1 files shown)


dervoeti (approved):
LGTM
We can think about making these failures visible in SecObserve in the future or moving this check to somewhere else (like our daily scanning of the release images). Not urgent though, can be part of a future rework.

PR #614, opened 2026-07-21
fix(template): Make "Finished Build and Publish" gate reflect real build outcome

lfrancke:
## Problem

The `finished` job (`Finished Build and Publish`) is the single required status check for the build workflow. We did this so we don't have to list every step in between as required.

During our 26.7 release we saw a build that succeeded even though it shouldn't have: https://github.com/stackabletech/listener-operator/actions/runs/29825092091

## Fix

Github considers _skipped_ jobs as succeeded rather than failing the job. So we need to make sure our `finished` job always runs.

This PR changes it by introducing an `if: always()` to the job

- **`if: always()`** so the gate always runs and reports a real success/failure conclusion.
- **List every leaf job directly in `needs`** and fail on any `failure`/`cancelled` result. `publish-index-manifest` was previously only a *transitive* dependency (via `openshift-preflight-check`), so its failure would not surface in `needs.*.result` even with `always()`.
- `skipped` is tolerated on purpose, jobs skip legitimately on `merge_group` events, forks, and when `detect-changes` finds no relevant changes.

### DIFF
--- a/template/.github/workflows/build.yaml.j2
+++ b/template/.github/workflows/build.yaml.j2
@@ -372,13 +372,41 @@ jobs:
     # WARNING: Do not change the name unless you will also be changing the
     # Required Checks (in branch protections) in GitHub settings.
     name: Finished Build and Publish
+    # Run even when a dependency failed, was skipped or cancelled, so that this
+    # gate reflects the real outcome. Without `always()` a failed dependency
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (commented):
Has this been tested on one of the operator repos?

NickLarsenNZ (approved):
LGTM

PR #616, opened 2026-07-22
fix(template): Use product string in docs templating script

Techassi:
This was fixed manually in downstream operators during the work for release 26.7.0. This PR fixes it directly in templating.


siegfriedweber (approved):
LGTM

I tested the airflow-operator and the commons-operator.

PR #617, opened 2026-07-27
fix: retry script should catch stackablectl errors

razvan:
# Description

Currently, the auto retry script for integration testing ignores stackablectl errors and reports failed runs as succesful.

Example: https://testing.stackable.tech/job/spark-k8s-operator-it-custom/84/

This pr fixes that.


NickLarsenNZ (approved):
LGTM

PR #619, opened 2026-08-03
fix: Set correct user IDs for container images build by nix

sbernauer:
Part of https://github.com/stackabletech/issues/issues/645

I noticed this while working on https://github.com/stackabletech/operator-rs/pull/1205, as the bundle-builder was rejected by Kubernetes up as it wanted to run as root.

I though of templating the IDs, but putting it in config/retired_files.yaml also felt wrong

### DIFF
--- a/template/default.nix
+++ b/template/default.nix
@@ -133,6 +133,12 @@ rec {
     ${entrypoint} crd > $out
   '';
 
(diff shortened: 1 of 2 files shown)


siegfriedweber (approved):
LGTM

PR #622, opened 2026-08-03
feat: Generate the Artifact Hub CRD annotation from extra/crds.yaml

lfrancke:
This is derived from extra/crds.yaml and ends up in Chart.yaml.

### DIFF
--- a/template/Makefile.j2
+++ b/template/Makefile.j2
@@ -28,7 +28,7 @@ docker-build:
 	docker build --force-rm --build-arg VERSION=${VERSION} -t "${OCI_REGISTRY_HOSTNAME}/${OCI_REGISTRY_PROJECT_IMAGES}/${OPERATOR_NAME}:${VERSION}-${ARCH}" -f docker/Dockerfile .
 
 ## Chart related targets
-compile-chart: version crds
+compile-chart: version crds crd-annotation
 
 chart-clean:
 	rm -rf "deploy/helm/${OPERATOR_NAME}/crds"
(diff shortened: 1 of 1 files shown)


template/Makefile.j2
```diff
@@ -43,6 +43,29 @@ crds:
...
+#       name: secretclasses.secrets.stackable.tech
+#       displayName: SecretClass
+#       description: A SecretClass is a cluster-global Kubernetes resource that defines ...
+#     - kind: TrustStore
+#       version: v1alpha1
+#       name: truststores.secrets.stackable.tech
+#       displayName: TrustStore
+#       description: A TrustStore requests information about how to validate secrets ...
+crd-annotation: crds
+	@CRDS="$$(yq ea -o=yaml '[.] | map(.spec.versions |= map(select(.storage == true))) | map({"kind": .spec.names.kind, "version": .spec.versions[0].name, "name": .metadata.name, "displayName": .spec.names.kind, "description": (.spec.versions[0].schema.openAPIV3Schema.description // "")})' extra/crds.yaml)" \
```
NickLarsenNZ: Does this overrwrite the extra/crds.yaml in the repo, or does it do something else that only gets seen when doing artifact hub stuff?

lfrancke: Instead of answering here I'll try to clarify the comment. Thanks for taking a look!

lfrancke: See if [this](https://github.com/stackabletech/operator-templating/pull/622/changes/85a78447a890ac2babaca386b7f1ced45965cecd) makes more sense now.

template/Makefile.j2
```diff
@@ -43,6 +43,40 @@ crds:
 	mkdir -p extra
 	cargo run --bin stackable-"${OPERATOR_NAME}" -- crd > extra/crds.yaml
 
+# This adds CRD metadata for artifacthub.io (AH) to Chart.yaml.
```
Techassi, line 46: I feel like this should be an xtask (or even a subcommand on the operator itself, like the `crd` subcommand) instead.

Also, this would need to be executed by pre-commit/prek to ensure consistency.

lfrancke: I want to push back on both.

1) We don't have xtask or anything like that for any other command in here and we already even do rewriting (for version stuff) with yq. So this is consistent

2) This is already executed via prek (regenerate-charts) so that's covered and it triggers when it's inconsistent.

template/Makefile.j2
NickLarsenNZ: Ah ok, so take some stuff from the generated CRDs, and make annotations in Chart.yaml.

I assume the changes were also tested on a real repo.

NickLarsenNZ (approved):
LGTM.

The yq statements could b easier to read on multiple lines but that can be done later (ie: when it bugs me enough, I'll do it).

template/Makefile.j2
lfrancke: I ran test.sh to verify the changes. 
Thanks.

template/Makefile.j2
lfrancke: @Techassi Can we resolve this or do you want to push back on my pushback?

Techassi: It is fine for now, but I still believe we should do this stricter environment (Rust) instead of using make+bash kung fo.

PR #624, opened 2026-08-04
feat: Add artifacthub-repo.yml for verified publisher and version filtering

lfrancke:
Artifact Hub reads this file from a special `artifacthub.io` tag on the chart's OCI repository, not from the chart itself. repositoryID enables the Verified publisher label.
And using the "ignore" thing we can keep dev, rc and PR tags out of AH.

### DIFF
--- a/template/deploy/helm/artifacthub-repo.yml.j2
+++ b/template/deploy/helm/artifacthub-repo.yml.j2
@@ -1,0 +1,25 @@
+---
+# Artifact Hub (AH) repository metadata.
+#
+# This is NOT part of the chart.
+# Artifact Hub reads it from a separate OCI artifact pushed to the chart repository under the special `artifacthub.io` tag:
+#
+#   oras push oci.stackable.tech/sdp-charts/{[ operator.name }]:artifacthub.io \
+#     --config /dev/null:application/vnd.cncf.artifacthub.config.v1+yaml \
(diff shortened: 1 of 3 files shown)


template/deploy/helm/artifacthub-repo.yml.j2
```diff
@@ -0,0 +1,25 @@
...
+# Every operator that gets this file has an ID. An operator without an Artifact Hub
+# repository excludes the file entirely via `ignored_files` in
+# config/repositories.yaml, the way druid does while it is being removed from SDP.
+repositoryID: {[ operator.artifacthub_repository_id }]
+
+# Versions Artifact Hub should not index.
+# This excludes all dev/rc/pr charts.
+ignore:
+  - name: {[ operator.name }]
+    version: '-(dev|rc|pr)'
```
NickLarsenNZ, line 25: Just checking, are we sure this doesn't do whole string matching? As in, they don't wrap that into a `^...$`?

lfrancke: https://artifacthub.io/packages/helm/stackable-listener-operator/listener-operator

It seems to work, yes.

NickLarsenNZ (approved):

template/deploy/helm/artifacthub-repo.yml.j2
Techassi: @NickLarsenNZ said I'm free to merge this. Marking as resolved.

PR #629, opened 2026-08-11
ci(template): Bump actions, add scheduled integration test

Techassi:
> [!CAUTION]
> The `schedule` interu profile needs to be present/commented in before the schedule will run.

This PR bumps all `stackabletech/actions` actions. It also enables the integration test workflow to run on a schedule.


template/.github/workflows/integration-test-dispatch.yml
```diff
@@ -1,10 +1,7 @@
 ---
-name: Integration Test
+name: Integration Test (Dispatch)
```
NickLarsenNZ: As mentioned, we should split this into a profile based workflow, and a kuttl based workflow (to set specific arguments like `--test` and `--parallel`)

template/.github/workflows/integration-test-schedule.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: Integration Test (Schedule)
+
+on:
+  schedule:
+    # At 00:00 on Sunday. See: https://crontab.guru/#0_0_*_*_0
+    - cron: "0 0 * * 0"
```
NickLarsenNZ: I also wonder if we want to stagger these eventually. Less load on the image registry, etc...

Wish there was a random delay option in Actions.

NickLarsenNZ (dismissed):
LGTM

template/.github/workflows/integration-test-schedule.yml
NickLarsenNZ: One idea is that we can make the schedule daily. Then each operator can have a local file with the day(s) that it should run. The workflow can bail out early if $today doesn't match the file.

Techassi: Yeah a `jitter` option would be nice. Also, it seems like we might want to shift the schedule slightly, because 00:00 UTC is very busy according to some threads online and results in massive delays.

template/.github/workflows/integration-test-dispatch.yml
Techassi: You can still set `--test` and `--test-suite` with the current approach. Getting rid of interu for the dispatch would also get rid of a whole bunch of wiring (like which cluster to use, setting up env vars, etc).

NickLarsenNZ: My point is the current list of inputs is kinda painful (and that's coming from me who knows how to use it).

I think others in the team will have similar complaints.

NickLarsenNZ: One idea I had was a prek hook that can update a drop-down in the action so it always has selectable items. That would remove some friction.

template/.github/workflows/integration-test-schedule.yml
NickLarsenNZ: Yeah makes sense. But I think staggering tests over different days would be better.

If there are problems it gives developers some space to fix things without being overloaded at the start of the week.

Techassi: > One idea is that we can make the schedule daily. Then each operator can have a local file with the day(s) that it should run. The workflow can bail out early if $today doesn't match the file.

We could leverage `interu` for that. Basically let it decide if the test should be run or not. We could model this via exit-codes or more structured data.

template/.github/workflows/integration-test-dispatch.yml
Techassi: Yeah that is a good idea.

Techassi: So I reworked the overall structure of the workflows in 0040114.

I did not yet add a script to generate inputs (which is technically not possible, because these workflows are templated but the inputs depend on repo-local values).

template/.github/workflows/integration-test-schedule.yml
Techassi: I pushed the scheduled time to 04:00 UTC on Sunday. The jitter/scatter is not implemented yet.

NickLarsenNZ (dismissed):
LGTM, but we should test before merging.

NickLarsenNZ (approved):

PR #632, opened 2026-08-14
ci(template): Change operator version behaviour

Techassi:
Previously, we only used 0.0.0-prXXX for the image tag if the PR targeted the main branch. This could lead to surprising results, because some PRs might be raised against a different feature branch (potentially part of a different PR). In these cases, the current version (most likely 0.0.0-dev) would be used and the -prXXX was appended to it, resulting in a surprising version tag 0.0.0-dev-prXXX.

To address this, we decided to only use the current version for PRs raised against release branches where this version scheme makes sense (that behaviour was also our intend when we initially wrote that script). For ANY other branch, we now always use 0.0.0-prXXX.

We could further improve the logic by checking if the current version differs from 0.0.0-dev and then use the version as is with a -prXXX suffix. If the version is 0.0.0-dev, we use the established 0.0.0-prXXX version.

### DIFF
--- a/template/.github/workflows/build.yaml.j2
+++ b/template/.github/workflows/build.yaml.j2
@@ -151,12 +151,18 @@ jobs:
 
           CURRENT_VERSION=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "stackable-{[ operator.name }]") | .version')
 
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (approved):
LGTM

PR #634, opened 2026-08-16
fix: Port the serviceAccount.name/create fix into the template

lfrancke:
Ports stackabletech/hive-operator#742 into the template. 

- the Deployment and the ServiceAccount resolve their name through `operator.serviceAccountName` instead of hardcoding `<fullname>-serviceaccount`, which is what makes `.Values.serviceAccount.name` take effect at all
- with `create=false` the helper fell back to `"default"`, giving a pod running as the namespace default ServiceAccount without the operator ClusterRole. It now requires `serviceAccount.name`.

This should not be rolled out before stackabletech/listener-operator#418 and stackabletech/secret-operator#736 are merged.

### DIFF
--- a/template/deploy/helm/[[operator]]/templates/_helpers.tpl
+++ b/template/deploy/helm/[[operator]]/templates/_helpers.tpl
@@ -65,9 +65,9 @@ Create the name of the service account to use
 */}}
 {{- define "operator.serviceAccountName" -}}
 {{- if .Values.serviceAccount.create }}
-{{- default (include "operator.fullname" .) .Values.serviceAccount.name }}
+{{- default (printf "%s-serviceaccount" (include "operator.fullname" .)) .Values.serviceAccount.name }}
 {{- else }}
-{{- default "default" .Values.serviceAccount.name }}
+{{- required "serviceAccount.name is required when serviceAccount.create is false, because the chart then does not create a ServiceAccount for the operator to run as." .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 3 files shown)


template/deploy/helm/[[operator]]/templates/_helpers.tpl
```diff
@@ -65,9 +65,9 @@ Create the name of the service account to use
 */}}
 {{- define "operator.serviceAccountName" -}}
 {{- if .Values.serviceAccount.create }}
-{{- default (include "operator.fullname" .) .Values.serviceAccount.name }}
+{{- default (printf "%s-serviceaccount" (include "operator.fullname" .)) .Values.serviceAccount.name }}
```
Comment (Techassi) [inline template/deploy/helm/[[operator]]/templates/_helpers.tpl:68]: Technically, we don't need to include `-serviceaccount` in the name, because the kind already indicates what the object is, a ServiceAccount.

We also do this in various other places, like the Deployments the operator creates. It wastes valuable characters and also duplicates information.

Comment (lfrancke) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:68]: That's true but that'd be a breaking change, no?
For now I just want to make it work, we can change it later/separately if needed.

Comment (NickLarsenNZ) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:68]: Yeah I'd like to remove the redundant kinds from the names in other objects too. But that can be done separately.

Comment (Techassi) [reply template/deploy/helm/[[operator]]/templates/_helpers.tpl:68]: Alright, I'm fine leaving it as is, but we should really get rid of the redundant data in names.

Techassi (approved):

PR #7, opened 2021-11-18
Add actions to build Helm charts to the operators and retire os package building.

soenkeliebau:
fixes #4 
fixes https://github.com/stackabletech/issues/issues/135

Tested with:
- [x] Druid
- [x] Superset
- [x] Hive
- [x] Trino
- [x] Zookeeper
- [x] Regorule
- [x] Opa
- [x] Kafka
- [x] Nifi
- [x] Spark
- [x] HDFS
- [x] HBase

PR #637, opened 2026-08-16
fix: Always create the operator ClusterRoleBinding

lfrancke:
FYI: Reported by @dervoeti on stackabletech/listener-operator#418

I actually saw this in hive-operator but thought there was a good reason for it I was missing. Seems there is not.
This was added in 2022: https://github.com/stackabletech/operator-templating/pull/7 but it actually came earlier in https://github.com/stackabletech/zookeeper-operator/pull/247 and Claude tells me that this looks very much like kubebuilder templates. I have not verified that but would trust it. So it seems we bootstrapped parts of that from kubebuilder but back then actually three objects were behind the `if`: ServiceAccount, ClusterRoleBinding and ClusterRole. @soenkeliebau  then split the ClusterRole out and moving only that OUT of the `if`. I assume none of that was intentional.

@dervoeti suggested renaming all of this to `rbac.create` but then we should maybe move the ClusterRole back as well. I suggest we merge this as is though and we can always reconsider....

The one problem I have is that this _might_ want a changelog entry which we can't do using templating as far as I'm aware... but at least a release note should be added.

### DIFF
--- a/template/deploy/helm/[[operator]]/templates/serviceaccount.yaml.j2
+++ b/template/deploy/helm/[[operator]]/templates/serviceaccount.yaml.j2
@@ -10,9 +10,11 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-# This cluster role binding allows anyone in the "manager" group to read secrets in any namespace.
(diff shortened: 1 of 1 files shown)


soenkeliebau:
Yes .. sorry, I cannot remember, but I believe this was not an intentional change on my part.

dervoeti (approved):

PR #638, opened 2026-08-17
chore(template): Use .yaml extension for artifacthub-repo.yml

Techassi:
We generelly discourage using `.yml` for YAML files. Artifact Hub supports both according to their tests and code.

This rename also resulted in yamllint to pick up the file.

### DIFF
--- a/config/repositories.yaml
+++ b/config/repositories.yaml
@@ -47,7 +47,7 @@ repositories:
     # Druid is being deprecated and removed from SDP, so it has no
     # Artifact Hub repository and no metadata file.
     ignored_files:
-      - deploy/helm/artifacthub-repo.yml.j2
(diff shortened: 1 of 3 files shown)


lfrancke:
That's the file name required by ArtifactHub: https://artifacthub.io/docs/topics/repositories/#verified-publisher

I have not checked whether it also accepts `.yaml`

Techassi:
> I have not checked whether it also accepts .yaml

Yup, it does. Both via OCI and HTTP(S).

https://github.com/artifacthub/hub/blob/f90cbb4ee79a31eaa348dbb2868ffa9d2cbe15d8/internal/repo/manager.go#L417-L440

NickLarsenNZ (approved):

lfrancke (approved):

